PR #2, opened 2018-04-18
SUIT-8932 Replace cosmiconfig with rc for ini support

givehug:


lib/utils/testLauncherHelper.js
```diff
@@ -54,23 +54,14 @@ function promptPassword() {
...
-		rc: '.suitestrc',
-		js: 'suitestrc.js',
-		packageProp: false,
-		rcExtensions: true,
-		rcStrictJson: true,
-		sync: true,
-	}).load();
-
-	return result ? result.config : {};
+	return rc('suitest', {}, () => ({}));
```
andriichumak: We should consider actually providing arguments parser (3rd argument) and let RC lib do all the props merging. But that can wait till next release.

PR #3, opened 2018-04-18
SUIT-8939 Add error watchers on current process in envHelper

givehug:


givehug:
@andriichumak, I was thinking whether we really need it for processes other than child process spawned by test launcher. We could think of defining special env variable that would indicate that current process is part of test launcher spawn. I'll add it to index for no though.

PR #5, opened 2018-04-19
SUIT-8895 Fixed unit tests for testLauncher/index.test.js. 

JeKachan:
Updated webbsockets package. Fixed TypeErrors which occurs due tests running. Fixed displaying device execution results for SuitestLauncher.

Unfortunately test coverage still fails. It happens because nyc inside himself use "spawn-wrap" which created child procesess, and this library create temporary execution folders in OS user folder ("Ievgen Kozynskyi" in my case) and path, with spaces, to executable file not wrapped in double quotes and by this reasons tests coverage fails in several places.

https://github.com/istanbuljs/nyc/issues/738
https://github.com/istanbuljs/nyc/issues/784


givehug:
Just to clarify, nyc coverage fails on windows only when there is a space id dir name.

lib/testLauncher/SuitestLauncher.js
```diff
@@ -77,11 +77,11 @@ class SuitestLauncher {
 				// log all devices logs
 				Object.keys(devicesLogs).forEach(dId => log.deviceLines(devicesLogs[dId], dId));
```
givehug: This is quite critical fix

PR #6, opened 2018-04-19
Hotfix/suit 8959

givehug:
Close session should not call http request for each test launcher child process. It should be called in main process after all children are finished.


lib/testLauncher/SuitestLauncher.js
```diff
@@ -52,6 +52,7 @@ class SuitestLauncher {
 						shell: true,
 						env: {
 							...process.env,
+							[envVars.SUITEST_LAUNCHER_PROCESS]: true,
```
andriichumak: I wonder if `true` is a good value. Would it be converted to string upon transfer to Chile process? If so, maybe it's a good idea to use here some string value to avoid confusion.

PR #7, opened 2018-04-19
feature/SUIT-8754 complete pending unit test imp

JeKachan:


test/testLauncher/index.test.js
```diff
@@ -8,7 +8,7 @@ const texts = require('../../lib/texts');
 const testLauncherTest = path.resolve(__dirname, '../../lib/utils/testHelpers/testLauncherTest.js');
 const spawn = require('child_process').spawn;
 
-describe('suitest test launcher', function() {
+describe.skip('suitest test launcher', function() {
```
givehug: should not be skipped?

lib/utils/assertThrowsAsync.js
```diff
@@ -3,9 +3,10 @@ const assert = require('assert');
 /**
  * Helper for testing async functions. Utilizes original node assert.throws.
  * @param {Function} fn async function
- * @param {any} regExp error regExp
+ * @param {RegExp|Function} regExp error regExp
```
AntonSBilan: in 9.11 it can also be an object

PR #10, opened 2018-04-23
Hotfix/suit 8877 unify the way unit tests are written

AntonSBilan:
Unified unit tests per folder for invalid input


test/chains/cookieChain.test.js
```diff
@@ -186,19 +187,16 @@ describe('Cookie chain', () => {
...
-			assert.strictEqual(error.code, SuitestError.INVALID_INPUT, 'code');
-		}
-		try {
-			await cookie(1);
-		} catch (error) {
-			assert.ok(error, 'error');
-			assert.strictEqual(error.code, SuitestError.INVALID_INPUT, 'code');
-		}
+	it('should throw error in case of invalid input', async() => {
+		await assertThrowsAsync(cookie.bind(null, undefined), {
```
JeKachan: No needs to pass second parameter

test/chains/networkRequestChain.test.js
```diff
@@ -266,15 +267,17 @@ describe('Network request chain', () => {
 				},
 			},
 		});
+	});
 
-		assert.throws(() => toJSON({}), 'Throw if no comparator and strategy defined');
-		assert.throws(() => toJSON({wasMade: true}), 'Throw if no comparator defined');
-		assert.throws(() => toJSON({
+	it('should throw error in case of invalid input', async() => {
+		await testInputError(toJSON, [{}]);
```
JeKachan: I would prefer avoid async/await noise. Could you implement sync version of testInputError and use async version only there where it really needed?

lib/utils/assertThrowsAsync.js
```diff
@@ -4,8 +4,9 @@ const assert = require('assert');
  * Helper for testing async functions. Utilizes original node assert.throws.
  * @param {Function} fn async function
  * @param {RegExp|Function|Object} regExp error
+ * @param {string} error msg
```
andriichumak: @AntonSBilan I think the correct jsdoc would look like `* @param {string} msg message for the assertion`

PR #14, opened 2018-04-24
Added MIT license file and badge to readme file

andriichumak:
@SuitLollipop please confirm it's OK.


andriichumak:
Will add license with GitHub tools instead

PR #20, opened 2018-04-25
Account for "appRunning" error type in JS API lib

AntonSBilan:


lib/utils/socketChainHelper.js
```diff
@@ -53,6 +53,15 @@ const processServerResponse = toString => function processServerResponseHandler(
 		);
 	}
 
+	if (res.errorType === 'appRunning') {
+		let msg = 'App is still running';
+
+		if (data.type === 'press') {
+			msg += `\nMaximum amount of key presses ${data.repeat} reached. Condition was not satisfied`;
```
andriichumak: OK, from the code I assume the only possible case when `appRunning` error can occur is for line "Press [] until application exits"... Is that so?

Also, all texts must be moved to `texts.js` file in `lib` folder.

AntonSBilan: When `suitest.application().hasExited()` we have just `appRunning` error but in case of press e.g. 
``` suitest.assert.press(suitest.VRC.UP).until(
			suitest.application().hasExited()
		).repeat(5).interval(2000);
 ```
we also add `conditionNotSatisfied` error

Comment (andriichumak) [reply lib/utils/socketChainHelper.js]: OK, in that case we should take into account that "application has exited" is not the only chain that can be passed to "until". And "until" can be chained to other subjects then "press".

File: lib/utils/socketChainHelper.js [hunk 2]
```diff
@@ -53,6 +54,15 @@ const processServerResponse = toString => function processServerResponseHandler(
 		);
 	}
 
+	if (res.errorType === 'appRunning') {
```
givehug: This is not what was planned actually. On line 14 we check if line is `queryFailed`, and if it is, and line is `testLine`, we throw AssertionError on line 48. The issue was that `application().hasExited()` chain fails not with `queryFailed_` type of error, but with `appRunning`. So basically we should simply add `appRunning` error type to test on line 14, so it is thrown as AssertionError.

Maybe it would be better to rename `const queryFailed` to `const isAssertionError` so it is more clear. 

There is a separate function for AssertionError message handling if we want some additional text for `application` and/or `until` chains.

lib/utils/socketChainHelper.js
givehug: I think this `conditionNotSatisfied` errorType is not very accurate. It was added on FE for press line only, but was not updated when we added browser lines (click, send text etc) (possible FE bug). If we really want this message `Maximum amount of key presses ${data.repeat} reached. Condition was not satisfied`, I suggest we simply add it to all failed testLine chains with `until` modifier and rephrase it to something like `Maximum amount of iterations was made. Condition was not satisfied`?

lib/texts.js
```diff
@@ -109,4 +109,8 @@ ${leaves}`,
 	'tl.finishedWithErrors': () => 'Some executions finished with errors',
 	'tl.finishedWithSuccess': () => 'All executions finished successfully',
 	'tl.promptPassword': () => 'Please enter your passwod: ',
+
+	// error messages
+	'err.msg.appRunning': 'App is still running',
```
givehug: we usually do it as arrow function

lib/texts.js
```diff
@@ -109,4 +109,8 @@ ${leaves}`,
...
+	'err.msg.conditionNotSatisfied': count => `Maximum amount of key presses ${count} reached. Condition was not satisfied`,
```
givehug: we usually do it as `template`

andriichumak: I think @JeKachan used this syntax already. It's OK with me as well as template function.

lib/utils/socketChainHelper.js
andriichumak: @AntonSBilan  please update according to @givehug comment.

PR #23, opened 2018-04-26
Feature/suit 9003 Add tvOS element properties to JS API

JeKachan:


lib/constants/state.js
```diff
@@ -0,0 +1,13 @@
+const state = {
```
andriichumak: It's not clear from the name what this variable is about. Please, make naming more specific.

index.d.ts
```diff
@@ -67,6 +71,10 @@ declare namespace suitest {
 		VRC: VrcConstants;
 		VIDEO_STATE: VideoState;
 		VISIBILITY_STATE: VisibilityState;
+		CONTENT_MODE: ContentMode;
+		STATE: State;
```
andriichumak: Naming should be more specific for state.

lib/constants/state.js
JeKachan: tvOSElementState?

andriichumak: How about `elementState`?

andriichumak: Also, we should keep it consistent with other constants. E.g. there is `ELEMENT_PROPS` (not `elementProps`). Please update new constants to fit that naming convention

PR #24, opened 2018-04-26
Fix misleading error messages in JS API when device is not available

AntonSBilan:


AntonSBilan:
Locally all tests passing

andriichumak:
@AntonSBilan the one that fails is "should log successful result for child process". Do you see cirlceci report or should I add some rights for you?

AntonSBilan:
@andriichumak I checked it and locally it passing, it was a problem recently which was fixed, maybe it is a cache, but I do not have permissions to run cirlceci one more time and without cache

lib/texts.js
```diff
@@ -104,6 +104,9 @@ Did you forget an "await"? If that's by design, make sure you call .abandon() af
 List of non-executed chains:
 ${leaves}`,
 
+	// ws
+	'ws.err.deviceProblem': error => `${error} \nThere is might be some problem with device \nPlease check if it is available and boot sequence is set up correctly`,
```
andriichumak: Please drop the first sentence here, it should start with "Please check...".

lib/api/webSockets.js
```diff
@@ -70,6 +70,10 @@ function handleResponse(msg) {
 			}); // resolve chain requests
 		} else if (res.result === 'success') {
 			req.resolve(res);
+		} else if (res.error && res.error.match(/device/igm)) {
```
andriichumak: I don't like detecting error type by searching keywords. How does `res` look? Is there some code? There should be at least a definite way to separate Suitest-produced error from network error.

AntonSBilan: @andriichumak  response is
```
{
  "messageId": "50b31530-4938-11e8-a6b0-15fec0ed97af",
  "content": {
    "result": "error",
    "error": "Device is disabled"
  }
}
```

andriichumak: OK, then make it SuitestError.UNKNOWN_ERROR with message from "Device is disabled".
WS error must appear only when network problem occurs. E.g. on "error" event in webSocket.js file.

lib/api/webSockets.js
```diff
@@ -70,6 +70,10 @@ function handleResponse(msg) {
...
+			// res.error can be Device is disabled, No device is available for test running ...
+			// some problem with device reject with appropriate msg
+			req.reject(new SuitestError(texts['ws.err.deviceProblem'](res.error), SuitestError.WS_ERROR));
 		} else {
 			req.reject(new SuitestError(texts.wsNotConnected() + (res.error ? ` ${res.error}.` : ''), SuitestError.WS_ERROR));
```
givehug: I think this line is wrong in general. We are throwing `WebSocket connection with Suitest server is not established` on any server response with error. Shouldn't we only display `texts.wsNotConnected()` on `ws.error`? And for any server response with error we could just use message from `content.error`.

andriichumak:
@AntonSBilan  please check pipelines. Does it fail due to UT instability or there is actually wrong UTs.

AntonSBilan:
@andriichumak there some changes required to ut, due to changes to websockets.js, I'm on it, please do not merge it yet

PR #25, opened 2018-04-27
replace real calls with mocks

AntonSBilan:


andriichumak:
@AntonSBilan I understand this mocks are for test launcher only, right? What about lib itself? Can we drop server there as well?

AntonSBilan:
@andriichumak Mocks used everywhere, we can drop testServer or replace it with mockedTestServer, I left it just in case we will need it in future

andriichumak:
@AntonSBilan please drop / replace it completely.

PR #27, opened 2018-05-02
Improve test launcher device logs

AntonSBilan:


lib/testLauncher/SuitestLauncher.js
```diff
@@ -68,8 +78,12 @@ class SuitestLauncher {
 							code,
 						}));
 
-					followChildProcess(child, line => {
-						devicesLogs[device.deviceId].push(line);
+					followChildProcess(child, (msg, type) => {
+						ws && ws.write('\n' + msg);
```
JeKachan: @andriichumak @givehug should we use platform depended line separators (os.EOL) instead of '\n'?

lib/testLauncher/SuitestLauncher.js
```diff
@@ -59,7 +59,17 @@ class SuitestLauncher {
 						},
 					};
 
-					log.startOnDevice(device.deviceId, this.restArgs.join(' '));
+					let ws = null;
```
givehug: I would rename  `ws` to `writeStream` for readability.

lib/testLauncher/launcherLogger.js
```diff
@@ -37,7 +40,7 @@ const launcherLogger = {
 function deviceLines(devicesLines, deviceId) {
 	launcherLogger._('');
 	launcherLogger._(`Test started on device: ${deviceId.slice(0, 8)}`);
-	devicesLines.forEach(log => launcherLogger._(log, deviceId));
+	devicesLines.forEach((log) => launcherLogger[log.type](log.msg, deviceId));
```
JeKachan: why here brackets around log?

lib/testLauncher/launcherLogger.js
```diff
@@ -102,6 +114,39 @@ function createTestPack() {
...
+	const pathToFile = path + `/${deviceId}.log`;
+
+	if (fs.existsSync(pathToFile)) {
+		fs.unlinkSync(pathToFile);
+	}
+
+	return fs.createWriteStream(path + `/${deviceId}.log`, {flags: 'ax'});
+}
+
+function mkDirByPathSync(targetDir) {
```
givehug: I would extract this to utils folder, add some unit test if not fully covered.

lib/testLauncher/launcherLogger.js
```diff
@@ -102,6 +114,39 @@ function createTestPack() {
 	launcherLogger._(stColors.mild('Creating test pack run...'));
 }
 
+function createWriteStream(path, deviceId) {
+	const pathToFile = path + `/${deviceId}.log`;
```
JeKachan: please use path separators depended on platform or normalize paths via "path" module.

lib/testLauncher/SuitestLauncher.js
```diff
@@ -59,7 +59,17 @@ class SuitestLauncher {
...
+
+					try {
+						ws = log.startOnDevice(device.deviceId, this.restArgs.join(' '), logDir);
+					} catch (e) {
```
givehug: @andriichumak , @JeKachan, @AntonSBilan  I'm thinking whether we should exit process with 1 code if writing to logs file was unsuccessful or just log some warning?

lib/texts.js
```diff
@@ -109,6 +109,8 @@ ${leaves}`,
 	'tl.finishedWithErrors': () => 'Some executions finished with errors',
 	'tl.finishedWithSuccess': () => 'All executions finished successfully',
 	'tl.promptPassword': () => 'Please enter your passwod: ',
+	'tl.failedToCreateDir': () => 'Unknown error occurred while creating a new directory',
+	'tl.startLogRecording': (dir, device) => `Logging started on ${device}\nYou can find logs in ${dir}/${device}.log.`,
```
JeKachan: Please adjust this message for windows platform

lib/testLauncher/SuitestLauncher.js
JeKachan: I prefer warning message. Or we should specify another error code for this case (on my point of view we should avoid using 1 for all errors)

lib/testLauncher/SuitestLauncher.js
givehug: I'm not sure.

lib/testLauncher/launcherLogger.js
```diff
@@ -102,6 +114,39 @@ function createTestPack() {
...
+function mkDirByPathSync(targetDir) {
+	const sep = path.sep;
+	const initDir = path.isAbsolute(targetDir) ? sep : '';
+	const baseDir = '.';
+
+	targetDir.split(sep).reduce((parentDir, childDir) => {
+		const curDir = path.resolve(baseDir, parentDir, childDir);
+
+		try {
+			fs.mkdirSync(curDir);
```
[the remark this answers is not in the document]
givehug: 'es6-promisify' is a good idea, not sure it would make sense to make it async in this case.

lib/utils/testLauncherHelper.js
```diff
@@ -88,13 +88,13 @@ function followChildProcess(child, callback) {
 		input: child.stdout,
 		terminal: false,
 	}).on('line', line => {
-		callback && callback(line.toString());
```
JeKachan: you can use callback without additional checking if will write at the function beginning: callback  = callback || () => {},

lib/testLauncher/SuitestLauncher.js
andriichumak: It probably does not matter at the end, but sure, let's use os.EOL

lib/testLauncher/SuitestLauncher.js
andriichumak: Agree, otherwise it can be confused with webSockets

lib/testLauncher/SuitestLauncher.js
andriichumak: I think exiting with error is the correct behaviour. Error message should clearly state that tests were aborted due to lack of write permission in the logs folder.

lib/texts.js
andriichumak: I.e. replace "\n" with os.EOL and "/" with path.sep

lib/testLauncher/launcherLogger.js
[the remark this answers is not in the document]
andriichumak: @JeKachan I think in this case it's OK to use blocking operation. This is ran only on script start-up.

If we do convert it to async/await, we'll need to make sure it's wrapped to the promise chain. Furthermore, it must be the same promise as bootstrap process, otherwise we'll get race condition.

lib/utils/testLauncherHelper.js
andriichumak: @JeKachan I think it's the same, just a matter of taste.

What should be added, however, is jsdoc mark that parameter is optional:
` * @param {Function} [callback] callback function`

lib/testLauncher/SuitestLauncher.js
```diff
@@ -5,7 +5,7 @@ const {handleChildResult, handleLauncherError, followChildProcess} = require('..
 const envVars = require('../constants/enviroment');
 const suitest = require('../../index.js');
 const {validate, validators} = require('../validataion');
-const {snippets: log} = require('./launcherLogger');
+const {snippets: log, launcherLogger} = require('./launcherLogger');
 
 class SuitestLauncher {
```
andriichumak: It seems that `runSutomatedSession` and `runInteractiveSession` have a lot in common. Should be refactored to avoid code duplication. If it's not too much work - do it in scope of this PR, because you added some more duplicated code here. Otherwise, please create a ticket and assign to me.

lib/testLauncher/buildArgs.js
```diff
@@ -25,6 +25,10 @@ const addCommandsAndHelp = yrgs => {
 		.alias('help', 'h')
 		.version(version)
 		.alias('version', 'v')
+		.option('logDir', {
+			describe: 'If defined, additional logging to specified folder will be performed',
```
andriichumak: Move text to `texts.js`

andriichumak:
Changes are OK now, please check UT

PR #30, opened 2018-05-03
SUIT-9121 Update startTest command

givehug:
Updated startTest command. clientTestId should be mandatory and name should become optional.

### DIFF
--- a/lib/commands/startTest.js
+++ b/lib/commands/startTest.js
@@ -12,19 +12,18 @@ const chainPromise = require('../utils/chainPromise');
 const {invalidInputMessage} = require('../texts');
 
 /**
- * Start automated session test, stop previous test, set test context.
- * @param {string} name - name of the test, must be unique per test pack
- * @param {Object} data - additional data
- * @param {string} data.clientTestId - user-defined id of the test, must be unique per test pack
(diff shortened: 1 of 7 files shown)


lib/commands/startTest.js
```diff
@@ -12,19 +12,18 @@ const chainPromise = require('../utils/chainPromise');
...
- * @param {string} data.clientTestId - user-defined id of the test, must be unique per test pack
- * @param {string} data.description - arbitrary description of the test
- * @param {string} data.appVersion - version of the application in test
- * @param {string} data.commitHash - other way to define version in test
+ * Start test
+ * @param {string} clientTestId - user-defined id of the test, must be unique per test pack
+ * @param {string} [name] - human readable name of the test, optional.
+ * @param {string} [description] - arbitrary description of the test
+ * @throws {SuitestError} - when context of the test run is wrong
  * @returns {ChainablePromise.<void>}
```
JeKachan: I think return type should be ChainablePromise.<void|SuitestError>

JeKachan:
@givehug Why yarn.lock was updated?

givehug:
@JeKachan, updated it when updating local packages, will revert.

PR #32, opened 2018-05-03
Fix a bug when promise chain created cycle dependency

andriichumak:
When used with Promise API instead of async/await, websocket-based commands would create cycle dependency and would never resolve. Now it's fixed.

Also, switched SuitestLauncher commands to unchained versions in scope of this ticket, because it was hard to debug without this changes.


lib/testLauncher/SuitestLauncher.js
```diff
@@ -26,64 +28,71 @@ class SuitestLauncher {
 		try {
```
JeKachan: do we really need this try-catch? Almost all covered code will be executed inside child processes.

andriichumak: There are also calls to server and we have to show to user a meaningful message in that case.
Second try/catch is only for "finally" part - to make sure session is closed on server properly if something goes wrong.

JeKachan: what if closeSessionUnchained on 42 line will failed? We will get finally block and there closeSessionUnchained will be called twice.

JeKachan: Sorry, I was a little mistaken, I talked about try catch on 37 line

andriichumak: You're right, should have put second try somewhere after line 45. Will update.

PR #36, opened 2018-05-04
Feature/suit 9119 unify configs

givehug:
- override lib config with .suitestrc user settings
- useSentry and logLevel can also be passed as cli args when using test launcher
- display deprecation warning for .configure() command


config/index.js
```diff
@@ -27,10 +36,17 @@ const config = {...(global._suitestTesting ? test : main)};
...
+function override(overrideObj = {}) {
+	const _overrideObj = pick(overridableFields, overrideObj);
+
+	validate(validators.CONFIGURE, _overrideObj, invalidConfigObj());
+
+	Object.assign(config, _overrideObj);
 }
 
+// override lib config with user config from .suitestrc
+override(readRcConfig());
```
andriichumak: @givehug This would actually mean that e.g. tokenKey will be read from rc file not only by launcher, but as well by lib itself. Wouldn't it conflict with env variables provided by launcher? Which of them would take precedence?

givehug: These keys are only used by launcher to start session and create child process. Child processes (envHelper) only use env vars provided by test launcher.

andriichumak: It's a bit confusing to have all those variables mixed up. Basically, child process has tokenKey and tokenPassword while it's actually authenticated with deviceAccessToken from env variable... That doesn't look right. It would make sense to me to have config and user params still separated internally and provide better filtering on which data should go to config and which to user params. Parent/child process should both have same config, but params are different. E.g.:

```js
config = {apiUrl, sentryDsn, useSentry, wsUrl, logLevel};
launcherParams = {tokenKey, tokenPassword, testPackId, logDir, concurrency, username, password, orgId, deviceId, appConfigId, inspect, inspectBrk};
libParams = {sessionType, sessionToken, deviceId, appConfigId};
```

Seems like that could be achieved if you add a more strict filter on rc file when merging it to config. What do you think?

givehug: Agree. I'll separate params.

givehug:
Merged

PR #40, opened 2018-05-13
Feature/suit 9117 check api js debugging on windows

AntonSBilan:
Feature/suit 9117 check api js debugging on windows

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -35,7 +35,13 @@ Debug config can be like:
 }
 ```
 
-# Debugging test with chrome devtools
+*In case of windows `node_modules/.bin` folder contains executable files
+ that are not javaScript programs so they cannot be launched with `node`
+ interpreter instead one can use direct path to javaScript file, e.g. 
+ `node_modules/suitest-js-api/lib/testLauncher/index.js` and in case of 
+ `mocha` `node_modules/mocha/bin/_mocha`  
+
+# Debugging tests with chrome devtools
 
 1. Run `npm run debug`.
 2. Open `http://localhost:[port]/json/list`.
--- a/lib/testLauncher/SuitestLauncher.js
+++ b/lib/testLauncher/SuitestLauncher.js
@@ -15,6 +15,7 @@ const {closeSessionUnchained} = require('../commands/closeSession');
 const {startTestPackUnchained} = require('../commands/startTestPack');
 const {snippets: log, launcherLogger} = require('./launcherLogger');
 const ansiRegex = require('ansi-regex');
+const texts = require('../texts');
 
 class SuitestLauncher {
 	constructor(ownArgv = {}, restArgs = []) {
@@ -28,7 +29,11 @@ class SuitestLauncher {
 		log.createTestPack();
 		validateInput('AUTOMATED', this.ownArgv);
 		try {
-			const {tokenKey, tokenPassword, testPackId, logDir} = this.ownArgv;
+			const {tokenKey, tokenPassword, testPackId, logDir, inspect, inspectBrk} = this.ownArgv;
+
+			if (inspect || inspectBrk) {
+				throw new Error(texts['tl.inspectOnlyForInteractiveMode']());
+			}
 			const queue = new Queue(this.ownArgv.concurrency);
 			const accessToken = await startTestPackUnchained({
 				accessTokenKey: tokenKey,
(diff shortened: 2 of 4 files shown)


demo/README.md
```diff
@@ -35,7 +35,13 @@ Debug config can be like:
 }
 ```
 
-# Debugging test with chrome devtools
+*In case of windows `node_modules/.bin` folder contains executable files
+ that are not javaScript programs so they cannot be launched with `node`
+ interpreter instead one can use direct path to javaScript file, e.g. 
+ `node_modules/suitest-js-api/lib/testLauncher/index.js` and in case of
```
JeKachan: What if we create launcher.js file in root and it will imports /lib/testLauncher/index.js, so user can use shorter path node_modules/suitest-js-api/launcher.js ?

AntonSBilan: Maybe can it be better to move it to bin folder?, like it is done in all other projects

lib/testLauncher/SuitestLauncher.js
```diff
@@ -128,14 +133,14 @@ class SuitestLauncher {
...
-						if (this.restArgs[0] === process.execPath) {
-							this.restArgs.shift();
-						}
+						this.restArgs.shift();
 						cp.fork(this.restArgs[0], this.restArgs.slice(1), {
 							...options,
 							execArgv: [
 								`--inspect${inspectBrk ? '-brk' : ''}=${inspectBrk || inspect}`,
 							],
+							cwd: `${process.cwd()}`,
```
JeKachan: @AntonSBilan why process.cwd() wrapped?

AntonSBilan: Not sure how it works on windows environment, I thought about potential problem with white spaces in path

demo/README.md
JeKachan: Do you mean creating bash/powershell scripts which will run launcher?

lib/testLauncher/SuitestLauncher.js
JeKachan: I believe that will not converted

demo/README.md
JeKachan: I think it good idea

lib/testLauncher/SuitestLauncher.js
```diff
@@ -28,7 +29,11 @@ class SuitestLauncher {
 		log.createTestPack();
 		validateInput('AUTOMATED', this.ownArgv);
 		try {
-			const {tokenKey, tokenPassword, testPackId, logDir} = this.ownArgv;
+			const {tokenKey, tokenPassword, testPackId, logDir, inspect, inspectBrk} = this.ownArgv;
+
+			if (inspect || inspectBrk) {
+				throw new Error(texts['tl.inspectOnlyForInteractiveMode']());
```
givehug: I think SuitestError should be used, perhaps of INVALID_INPUT type.

AntonSBilan: In case of SuitestError output in console will be rather big with call stack, not sure that it has sense to show this info to user, If you do not like using just Error probably it is better to rewrite it to avoid using Error at all?

givehug: I think when we throw Error explicitly it should be SuitestError, so its clear that it was defined by our library. (AssertionError for assertions). In this case if using inspect || inspectBrk doesn't break anything, maybe its better to just display a warning with logger.warn? @andriichumak

bin/suitest
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
```
JeKachan: This line really need?

demo/README.md
```diff
@@ -35,7 +35,13 @@ Debug config can be like:
 }
 ```
 
-# Debugging test with chrome devtools
+*In case of windows `node_modules/.bin` folder contains executable files
+ that are not javaScript programs so they cannot be launched with `node`
+ interpreter instead one can use direct path to javaScript file, e.g. 
+ `node_modules/suitest-js-api/bin/suitest` and in case of
```
JeKachan: @AntonSBilan @givehug @andriichumak  If we have have configuration which works on all platforms, maybe better use it and dont separate guide for different platforms?

bin/suitest
AntonSBilan: Yes, at least on mac .bin folder contains plain js files and we need it to make it executable not sure about windows but it may also use it to create appropriate .cmd file

demo/README.md
AntonSBilan: As for me I prefer to use files in ./bin folder, no need to look for correct path in package.json but @andriichumak  has it sense to make separate section for windows users with example configs in readme and user docs or these who use js api are rather advanced to know about their platform specific issues?

andriichumak: I think it's enough as long as we account for things that user can't fix on his end. E.g. make sure that **all** string parameters are wrapped with quotes when needed (i.e. when they hold spaces).

PR #43, opened 2018-05-16
SUIT-9280 Add queryTimeout to the list of AssertionErrors

JeKachan:


lib/utils/socketChainHelper.js
```diff
@@ -11,7 +11,8 @@ const texts = require('../texts');
  * @param {Function} toString
  */
 const processServerResponse = toString => function processServerResponseHandler(res, data) {
-	const isAssertionError = res.errorType && !!res.errorType.match(/^queryFailed|^appRunning|^appNotRunning/);
+	const isAssertionError = res.errorType &&
+		!!res.errorType.match(/^queryFailed|^appRunning|^appNotRunning|^queryTimeout/);
```
andriichumak: I wonder if it would make more sense to do `[].includes` instead of RegExp.

givehug: There are several types of `queryFailed` errors `queryFailed_...something` , test takes it into account. All other errors in regex are constant as they are here, I believe.

andriichumak: ah, OK, makes sense now.

JeKachan: `queryFailed_...something` - this error specific only for FE module. That module compose `errorType` and message.code in this way: `${errorType}_{message.code}` and this means that `queryFailed` have some subtype. So, I think we can replace this regex to .includes()

PR #44, opened 2018-05-17
SUIT-9274 toString methods of all chains must account for negation

JeKachan:


andriichumak:
@JeKachan we agreed earlier that we should support mocha-style negations, i.e. switch expected and actual values in AssertionError. Would be weird to use different messages in one case and same messages with switched expected/actual in other cases. Would it make sense to normalise it throughout the app?

PR #45, opened 2018-05-18
SUIT-9277 Make sure `match` chain method accepts Symbol constants for enumerable values

AntonSBilan:
SUIT-9277  Make sure `match` chain method accepts Symbol constants for enumerable values


lib/validataion/jsonSchemas.js
```diff
@@ -201,6 +206,35 @@ schemas[validationKeys.ELEMENT_PROPS] = {
 				'val': {'type': 'boolean'},
 				deviation,
 			}},
+			{'properties': {
+				'name': {'const': ELEMENT_PROP.BORDER_STYLE},
+				'type': propComarator,
```
givehug: 'val' missing ?

AntonSBilan: We have only apple border style values, maybe it has sense to add all possible border-style values so it will be possible to validate it with 'enum'

andriichumak: I agree. @AntonSBilan, could you please add it in scope of this PR?

andriichumak: Not only for border-style, but for other enumerable properties as well. Also, please update our user docs and Confluence specs to include this new constants to the list.

lib/chains/elementChain.js
```diff
@@ -158,7 +158,7 @@ const toJSON = data => {
 				if (prop.val === VALUE.REPO) {
 					out.inherited = true;
 				} else {
-					out.val = prop.val;
+					out.val = typeof prop.val === 'symbol' ? ELEMENT_VALUES[prop.name][prop.val] : prop.val;
```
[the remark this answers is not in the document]
AntonSBilan: Not sure if I understand correctly, but currently ELEMENT_VALUES is same as other maps in mappings.js, string in this case is any user input witch is not of type enum like `color` or `width`

andriichumak:
@JeKachan could you please review latest changes

lib/chains/elementChain.js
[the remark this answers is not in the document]
givehug: So are strings allowed ? or symbols only?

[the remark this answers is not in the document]
AntonSBilan: Only symbols, changed just for consistency

PR #47, opened 2018-05-21
SUIT-9312 Add negation composer for network request chain

JeKachan:
Added notComposer.
Changed toString.
Changed type definition.


test/chains/networkRequestChain.test.js
```diff
@@ -119,6 +165,31 @@ describe('Network request chain', () => {
...
+				timeout: 2000,
+				condition: {
+					subject: {
+						type: 'network',
+						requestInfo: [],
+						responseInfo: [],
+						compare: '!=',
+						val: 'test',
+					},
+					type: 'made',
```
andriichumak: I'm not sure `subject.compare` is the right place to apply negation. Shouldn't it be applied to `type`? I.e. `!made`.

andriichumak: @polomsky could you confirm which option is correct?

polomsky: you are right, only `made` and `!made`, `!=` can be part of `requestInfo` or `responseInfo`

PR #49, opened 2018-05-21
SUIT-9168 Add videoAdapter error message handling

givehug:
Add Instrumentation Library video adapter related error messages


lib/utils/socketChainHelper.js
```diff
@@ -127,6 +128,22 @@ function getAssertErrorMessage(res, text, data) {
...
+/**
+ * Create human readable general error message from suitest error response
+ * @param {Object} res
+ * @returns {string}
+ */
+function getGeneralErrorMessage(res) {
+	const errorType = path(['result', 'errorType'], res);
+	const errorReason = path(['result', 'message', 'info', 'reason'], res) || {};
+
+	if (errorType === 'deviceError' && videoAdapterErrors.includes(errorReason.code)) {
```
JeKachan: if I remember right, `reason` has string type?

lib/utils/socketChainHelper.js
```diff
@@ -127,6 +128,22 @@ function getAssertErrorMessage(res, text, data) {
...
 
+/**
+ * Create human readable general error message from suitest error response
+ * @param {Object} res
+ * @returns {string}
+ */
+function getGeneralErrorMessage(res) {
+	const errorType = path(['result', 'errorType'], res);
+	const errorReason = path(['result', 'message', 'info', 'reason'], res) || {};
+
```
givehug: @andriichumak , we changed reason type of IL (errorHandler ret.errorType) for video adapter errors only from string to object with code and message. I think it would be better if we keep it as string and add another key for message `errorText` etc, if no changes from BE are required.

andriichumak: Yeah, I changed that in IL in that way specifically to avoid extra work on BE. I'll discuss that with Tomas.

givehug:
going to rewrite this after error messages hotfix

PR #51, opened 2018-05-21
SUIT-9292 Change error message for index property in element selector of JS API

AntonSBilan:
SUIT-9292
Change error message for index property in element selector of JS API


AntonSBilan:
@andriichumak Run test couple times error did not occur

andriichumak:
@AntonSBilan please check what's wrong with Unit Tests

andriichumak:
There seems to be a race condition in UTs. Re-running them. We should look into that during next maintenance.

PR #52, opened 2018-05-22
SUIT-9302 Update JS API VR buttons

AntonSBilan:
SUIT-9302 Update JS API VR buttons


lib/mappings.js
```diff
@@ -56,10 +56,14 @@ VRC[VRC[vrcConstants.CH_UP] = 'CH_UP'] = vrcConstants.CH_UP;
 VRC[VRC[vrcConstants.CH_DOWN] = 'CH_DOWN'] = vrcConstants.CH_DOWN;
 VRC[VRC[vrcConstants.HOME] = 'HOME'] = vrcConstants.HOME;
 // xBox-specific
-VRC[VRC[vrcConstants.A] = 'ENTER'] = vrcConstants.A;
-VRC[VRC[vrcConstants.B] = 'BACK'] = vrcConstants.B;
-VRC[VRC[vrcConstants.X] = 'BLUE'] = vrcConstants.X;
-VRC[VRC[vrcConstants.Y] = 'YELLOW'] = vrcConstants.Y;
+VRC['A'] = vrcConstants.A;
```
JeKachan: @AntonSBilan why A, B, X, Y constants wrote in different way than others?

AntonSBilan: according to Taras comment to corresponding parent ticket` id "BUTTON_A" should be added with the name "A"`

JeKachan: Unfortunately, this is little bit tricky for me.
For exmple: VRC[vrcConstants.DPAD_UP] === 'DPAD_UP' so VRC['DPAD_UP'] === vrcConstants.DPAD_UP
But for VRC['A'] === vrcConstants.A and VRC[vrcConstants.A] !== 'A'?. This something undetectable, because VRC designed to be as enum, but this changes broke this design. Could you please add comment with details why we need this for avoid misunderstanding in future. Maybe needed add additional constants for split default vrc buttons from xbox?

andriichumak: See Jira ticket

andriichumak:
@JeKachan please review latest changes.

PR #53, opened 2018-05-22
SUIT-9356 Move mucha set-up demo from JS API repo to demo repo

AntonSBilan:
SUIT-9356  Move mucha set-up demo from JS API repo to demo repo


.gitignore
```diff
@@ -56,7 +56,7 @@ typings/
 # dotenv environment variables file
 .env
 
-demo/testConfig.js
+demo/
```
andriichumak: Shouldn't we drop demo/ completely?

PR #54, opened 2018-05-23
report about apple support

polomsky:


andriichumak:
@polomsky, this will go the next release. Probably tomorrow.

PR #56, opened 2018-05-24
SUIT-9282 Add verbose messages to JS API

AntonSBilan:
SUIT-9282 Add verbose messages to JS API


lib/composers/thenComposer.js
```diff
@@ -18,6 +19,16 @@ const makeThenComposer = (getSocketMessage, callback) => makeMethodComposer(
 			} else {
 				const socketMessage = getSocketMessage(data);
 
+				if (['query', 'testLine', 'eval'].includes(socketMessage.type)) {
```
givehug: This expression `['query', 'testLine', 'eval'].includes`  is also used in websockets module, could be extracted as function.

andriichumak: I don't like the idea of thenComposer knowing about socket message types. We are mixing two different concerns here.
Let's keep code clean and instead provide to `makeThenComposer` a `beforeSend` callback. This way each chain can decide if it wants to log a message and what message should it be. As a bonus, there would be no need to pass `toString` and `methodToString` here.

lib/chains/cookieChain.js
```diff
@@ -52,6 +52,7 @@ const toString = data => {
 			);
 	}
 };
+const methodToString = (data) => `cookie(${data.cookieName || ''})`;
```
andriichumak: There is no need for safeguard (`|| ''` part), as cookie name goes through validation.

lib/chains/locationChain.js
```diff
@@ -54,10 +54,11 @@ const toJSON = data => {
 
 	return socketMessage;
 };
+const methodToString = () => 'location(...)';
```
andriichumak: Location does not accept any parameters

lib/chains/networkRequestChain.js
```diff
@@ -108,9 +108,10 @@ const toJSON = data => {
 
 	return socketMessage;
 };
+const methodToString = () => 'networkRequest(...)';
```
andriichumak: Network request does not accept any parameters

lib/chains/openAppChain.js
```diff
@@ -38,9 +38,10 @@ const toJSON = data => {
 
 	return socketMessage;
 };
+const methodToString = (data) => `openApp(${data.relativeURL || '...'})`;
```
andriichumak: No need to have `|| '...'` - there is a validation

lib/chains/openUrlChain.js
```diff
@@ -25,10 +25,11 @@ const toJSON = data => ({
 		url: data.absoluteURL,
 	},
 });
+const methodToString = (data) => `openUrl(${data.absoluteURL || '...'})`;
```
andriichumak: No need to have || '...' - there is a validation

lib/chains/pollUrlChain.js
```diff
@@ -16,6 +16,7 @@ const {
 const {pollUrl} = require('../texts');
 
 const toString = ({url, response}) => pollUrl(url, response);
+const methodToString = (data) => `pollUrl(${data.url || '...'})`;
```
andriichumak: For some reason it's not validated, however instead of adding `|| '...'` I would rather made a proper validation of input parameters.

lib/chains/positionChain.js
```diff
@@ -48,6 +48,7 @@ const toString = data => {
 
 	return out;
 };
+const methodToString = (data) => `position(${data.coordinates && JSON.stringify(data.coordinates) || '...'})`;
```
andriichumak: No need to have || '...' - there is a validation

lib/chains/pressButtonChain.js
```diff
@@ -43,6 +43,7 @@ const toString = data => {
 
 	return out;
 };
+const methodToString = (data) => `pressButton(${data.ids && mapIds(data.ids).join(', ') || '...'})`;
```
andriichumak: No need to have || '...' - there is a validation

lib/chains/windowChain.js
```diff
@@ -95,6 +95,7 @@ const getBrowserCommandType = data => {
 		return 'acceptAlertMessage';
 	}
 };
+const methodToString = () => 'window(...)';
```
andriichumak: Window does not accept any parameters

lib/chains/pollUrlChain.js
```diff
@@ -63,7 +66,7 @@ const getComposers = data => {
 
 const pollUrlChain = (url, response) => makeChain(getComposers, {
 	type: 'pollUrl',
-	url,
+	url: validate(validators.NON_EMPTY_STRING, url, invalidInputMessage('pollUrl', 'URL')),
 	response,
```
JeKachan: @andriichumak @givehug here should be some validation for `response`?

lib/utils/logger.js
```diff
@@ -106,3 +106,13 @@ Object.values(methods).forEach(method => {
 });
 
 module.exports = logger;
+
+module.exports.logVerboseMessage = (...args) => {
```
JeKachan: @AntonSBilan could you add JSDoc here, please?

lib/chains/pollUrlChain.js
andriichumak: Yes, it should be same - non-empty string.

lib/utils/logger.js
andriichumak: Why do you need this? According to logger util, `info` messages will be displayed when `config.logLevel` is set to `verbose || debug`. This wrapper seems unnecessary. Please, just use `logger.info`.

andriichumak:
Still some changes needed, described in Jira ticket

lib/testLauncher/SuitestLauncher.js
```diff
@@ -37,12 +37,17 @@ class SuitestLauncher {
 				process.exit(1);
 			}
 			const queue = new Queue(this.ownArgv.concurrency);
+			const consoleInfo = console.info;
```
andriichumak: I'm not particularly fond of this solution. We are using our own logger anyway. Isn't it better to make it aware of execution context?

I would make a single logger module, that would export either logger or launcherLogger depending on context. And this one module would be used everywhere in the process. Similar how in React app on Swiss project we have store and store.dev and choose one based on env variable.

AntonSBilan: ok

lib/testLauncher/SuitestLauncher.js
```diff
@@ -61,7 +62,8 @@ class SuitestLauncher {
 							shell: true,
 							env: {
 								...process.env,
-								[envVars.SUITEST_LAUNCHER_PROCESS]: true,
+								[envVars.SUITEST_LAUNCHER_MAIN_PROCESS]: false,
+								[envVars.SUITEST_LAUNCHER_CHILD_PROCESS]: true,
```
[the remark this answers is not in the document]
givehug: Maybe lets change SUITEST_LAUNCHER_PROCESS to type string it in scope of this ticket? Give it some constant values "main", "child".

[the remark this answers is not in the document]
AntonSBilan: there are 2 variables to distinguish Launcher from user script and main process from child (as 
for child process we logging in main process we need to avoid double '    >' prefixes ) @andriichumak what do you think about @givehug proposal (anyway env veriables are string and true false are strings as well)?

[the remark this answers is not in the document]
andriichumak: OK, let's have one variable with enumerated values. But make it **launcher** and **lib**, because our lib can be used without launcher as well. Also, make sure that **lib** is a default value when none provided.

lib/commands/closeSession.js
```diff
@@ -18,8 +19,9 @@ const {authNotAllowed} = require('../texts');
  * @returns {Promise} response object
  */
 async function closeSession() {
+	logger.info(commandWillBeExecuted('closeSession'));
 	// don't invalidate tokens in test launcher child processes
-	if (!process.env[envVars.SUITEST_LAUNCHER_PROCESS]) {
+	if (process.env[envVars.SUITEST_LAUNCHER_PROCESS] !== 'child') {
```
JeKachan: @AntonSBilan maybe make sense move 'child' and 'main' to constants

lib/utils/logger.js
```diff
@@ -1,13 +1,26 @@
...
+	bold: colors.cyan.bold,
+	errorColor: colors.red,
+	successColor: colors.green,
+};
 
 const methods = {
 	log: {
 		key: 'log',
 		color: 'white',
+		launcherColor: null,
```
JeKachan: If null dont have any special meaning, I would prefer replace it ti empty string

lib/utils/logger.js
```diff
@@ -83,26 +100,59 @@ const log = method => (...args) => {
...
+const launcherLogger = {
+	colors: stColors,
+};
+
+/**
+ * Prefix log with device id
+ * @param {string} devId
+ */
+function devicePrefix(devId) {
+	return devId ? 'Device ' + devId.slice(0, 8) + '> ' : '               > ';
```
JeKachan: How much spaces in value from false branch? You can use something like `'> '.padStart(10, ' ')`

andriichumak:
@AntonSBilan @givehug We'll need to refactor logging when we have time. For now this should do. That's why I did not cover 100% with UTs.

lib/utils/interactiveProgressHanlder.js
```diff
@@ -34,7 +34,7 @@ function getProgressExplanation(status, code = '') {
 function handleProgress(status, code) {
 	const text = getProgressExplanation(status, code);
 
-	text && console.log(stColors.suit(`  - ${text}`));
+	text && logger.info(text);
 }
```
givehug: Forgot dashes and spaces here ?
`  - ${text}`

lib/utils/logger.js
```diff
@@ -67,12 +94,11 @@ Object.freeze(methods);
  * Logger function.
  * @param {*} method
  */
-const log = method => (...args) => {
+const log = (method) => (...args) => {
```
givehug: Lets describe two args here (text, deviceId) for clarity, and pass text on line 122 instead of ...args, so first argument is always message text and second deviceId?

lib/utils/interactiveProgressHanlder.js
andriichumak: I removed this intentionally, because it's now displayed with "Info: " prefix.

givehug: should be mentioned in ticket I think, because that was request form qa team

andriichumak: ok. Will add a comment there

PR #57, opened 2018-05-25
SUIT-9273 Add "isNot" alias for not chain link

AntonSBilan:
SUIT-9273
Add "isNot" alias for not chain link

### DIFF
--- a/lib/composers/intervalComposer.js
+++ b/lib/composers/intervalComposer.js
@@ -4,7 +4,7 @@ const {validate, validators} = require('../validataion');
 const {invalidInputMessage} = require('../texts');
 
 /**
- * Defines not and doesNot methods
+ * Defines not and doesNot and isNot methods
  */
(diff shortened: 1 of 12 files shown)


lib/composers/intervalComposer.js
```diff
@@ -4,7 +4,7 @@ const {validate, validators} = require('../validataion');
 const {invalidInputMessage} = require('../texts');
 
 /**
- * Defines not and doesNot methods
+ * Defines not and doesNot and isNot methods
```
givehug: I think this line is copy pasted from notComposer 8), should be changed

lib/composers/repeatComposer.js
```diff
@@ -4,7 +4,7 @@ const {validate, validators} = require('../validataion');
 const {invalidInputMessage} = require('../texts');
 
 /**
- * Defines not and doesNot methods
+ * Defines not and doesNot and isNot methods
```
givehug: same here, copy pasted probably

PR #58, opened 2018-06-05
SUIT-6531/SUIT-9302 Update JS API VR buttons

JeKachan:


andriichumak:
@JeKachan it shouldn't target master, it's not a hotfix

PR #60, opened 2018-06-07
SUIT-9592 Update useSentry flag naming in JS API changed to disallowCrashReports

AntonSBilan:
SUIT-9592 Update useSentry flag naming in JS API changed to disallowCrashReports

### DIFF
--- a/lib/texts.js
+++ b/lib/texts.js
@@ -127,5 +127,5 @@ ${leaves}`,
 
 	// Cli args
 	cliLogLevel: () => 'Logger verbosity level',
-	cliUseSentry: () => 'Allow suitest to use Sentry error reporting',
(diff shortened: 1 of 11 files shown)


lib/texts.js
```diff
@@ -127,5 +127,5 @@ ${leaves}`,
 
 	// Cli args
 	cliLogLevel: () => 'Logger verbosity level',
-	cliUseSentry: () => 'Allow suitest to use Sentry error reporting',
+	cliDisallowCrashReports: () => 'Allow suitest to use Sentry error reporting',
```
andriichumak: Please, replace error message with "Disallow Suitest to submit automatic crash reports when unexpected exceptions occur"

andriichumak:
@AntonSBilan please fix UT

PR #62, opened 2018-06-11
SUIT-9626 Fix UT failing after merge with master.

JeKachan:
Fixed some kind of cyclic module referencing: testLaunchHelper.js <-> requests.js


JeKachan:
Changed text for SuitestError.UNKNOWN_ERROR according to confluence

PR #63, opened 2018-06-12
SUIT-9466 Add interactive test progress handler

givehug:
During an interactive test execution there are often periods of time where longer waits occur and it is not immediately obvious why it is so.
In this case server sends WS events with content type 'progress'.
Description is printed to stdout.


lib/texts.js
```diff
@@ -124,4 +124,29 @@ ${leaves}`,
 	'err.msg.appRunning': () => 'App is still running',
 	'err.msg.conditionNotSatisfied': count => `Maximum amount of key presses ${count} reached. Condition was not satisfied`,
 	'err.msg.appNotRunning': () => 'Application is not running',
+
+	// interactive test progress status
+	'interactiveProgress.status.openingApp': () => 'Trying to open app...',
+	'interactiveProgress.status.closingApp': () => 'Trying to open app...',
```
AntonSBilan: probably should be 'close app...'

lib/texts.js
```diff
@@ -124,4 +124,29 @@ ${leaves}`,
...
+	'interactiveProgress.code.candyBoxOffline': () => 'Cannot continue: CandyBox controlling this device is offline.',
+	'interactiveProgress.code.suitestDriveOffline': () => 'Cannot continue: SuitestDrive controlling this device is offline.',
+	'interactiveProgress.code.runningBootSequence': () => 'Trying to open Suitest bootstrap application.',
+	'interactiveProgress.code.deviceInUse': () => 'A user is currently connected to this device. Execution will continue after the user disconnects.',
+	'interactiveProgress.code.deviceDisabled': () => 'This device is disabled. For the execution to continue please enable the device.',
+	'interactiveProgress.code.deviceDeleted': () => 'Cannot continue: Device is deleted',
+	'interactiveProgress.code.internalError': () => 'Cannot continue: Internal error occurred.',
+	'interactiveProgress.code.notDefinedPlatform': () => 'Cannot continue: Device does not support this platform.',
+	'interactiveProgress.code.lgWebosPlatformError': () => 'Cannot continue: LG WebOS driver failed.',
+	'interactiveProgress.code.xboxPlatformError': () => 'Cannot continue: LG WebOS driver failed.',
```
AntonSBilan: same line as above, is it ok?

givehug: texts were copied from ticket, fixed without approval

PR #64, opened 2018-06-13
SUIT-9593 Make sure CLI params have higher priority then RC params

andriichumak:


lib/utils/testLauncherHelper.js
```diff
@@ -80,7 +80,7 @@ function readLauncherConfig() {
  */
```
givehug: Please update description, so its clear that configB is returned unless its property nullable

andriichumak: updated

PR #66, opened 2018-06-13
Feature/suit 9466 fix log format

givehug:


givehug:
Update colors version

PR #67, opened 2018-06-13
SUIT-9587 Add hint for command-specific help messages to JS API

AntonSBilan:
SUIT-9587
Add hint for command-specific help messages to JS API

### DIFF
--- a/lib/testLauncher/buildArgs.js
+++ b/lib/testLauncher/buildArgs.js
@@ -2,6 +2,7 @@ const yargs = require('yargs/yargs');
 const automatedCommand = require('./commands/automated');
 const interactiveCommand = require('./commands/interactive');
 const {version} = require('../../package.json');
+const {stColors} = require('./launcherLogger');
 
 /**
  * Build main argv
@@ -12,6 +13,7 @@ const buildYargs = () => {
 		.boolean('help')
 		.boolean('version')
 		.help(false)
+		.epilog(`Run ${stColors.bold('suitest interactive --help')} or ${stColors.bold('suitest automated --help')} to see those command's options`)
 		.version(false);
 };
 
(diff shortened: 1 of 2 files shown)


lib/testLauncher/buildArgs.js
```diff
@@ -12,6 +13,7 @@ const buildYargs = () => {
 		.boolean('help')
 		.boolean('version')
 		.help(false)
+		.epilog(`Run ${stColors.bold('suitest interactive --help')} or ${stColors.bold('suitest automated --help')} to see those command's options`)
```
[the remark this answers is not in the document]
andriichumak: It's OK. Are you sure it's not added when user runs other commands then --help?

andriichumak:
@AntonSBilan we found another issue that is somewhat related, would be great if you could fix it in scope of this one.
When user runs just `suitest` without options or command - it would output same as with `--help` argument. Note that `suitest -v` should still only output version, not help message.

lib/testLauncher/buildArgs.js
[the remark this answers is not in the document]
AntonSBilan: For me it only appears with suitest --help

lib/testLauncher/buildArgs.js
```diff
@@ -12,6 +14,8 @@ const buildYargs = () => {
 		.boolean('help')
 		.boolean('version')
 		.help(false)
+		.check((hash, argv) => !argv.length && stColors.errorColor(texts['tl.noArguments']()), false)
+		.epilog(`Run ${stColors.bold('suitest interactive --help')} or ${stColors.bold('suitest automated --help')} to see those command's options`)
```
andriichumak: Could you move the text to texts? There is a way to inject variables with tagged template.

andriichumak: Also, I think there is no need for apostrophe in "command's" after all, as it's plural.

lib/texts.js
```diff
@@ -121,6 +121,7 @@ ${leaves}`,
 
 	// test launcher
 	'tl.nodevices': () => 'there is no device in config to proceed with testing',
+	'tl.noArguments': () => 'There was no argument or command specified',
```
andriichumak: Command was not specified

PR #69, opened 2018-06-14
SUIT-9649 Errors from `getComposers` are not reported to Sentry

AntonSBilan:
SUIT-9649
Errors from `getComposers` are not reported to Sentry


andriichumak:
@AntonSBilan , can't you use https://mineus.atlassian.net/browse/SUIT-9581 ?

AntonSBilan:
@andriichumak Tried `Raven.wrap` but it does not report error, not sure why, but at least `shouldSendCallback` not fired

lib/utils/sentry/Raven.js
```diff
@@ -37,6 +37,20 @@ function captureException(error) {
 	return new Promise(res => Raven.captureException(error, res));
 }
 
+function wrap(func) {
```
andriichumak: Please, reformat function according to aslant rules

andriichumak:
@AntonSBilan Are you sure your version of works as expected? I was able to get one Sentry error, but then I changed something in code, removed error from Sentry logs and ran it again - was unable to receive any more reports. It seems that Setry's `captureException` is async and sometimes process ends before it's sent. Could you double-check it, please.

lib/utils/makeChain.js
```diff
@@ -1,6 +1,7 @@
 const {mergeAll} = require('ramda');
+const Raven = require('raven');
 
-const makeChain = (getComposers, data) => {
+const makeChain = Raven.wrap((getComposers, data) => {
```
givehug: It should work, I was able to send reports constantly. I think it matters where you put Raven.wrap( statement.

AntonSBilan: that's wired,  I tried it on my mac and it does not even get to 'shouldSendCallback' method (Raven creates new domain and run wrapped function and for me domain.on(error) not fired)

PR #72, opened 2018-06-14
Feature/suit 9648

givehug:
When receiving fatal errors from server node process running the test will be stopped with informative message.
New configuration variable in RC file added, "dieOnFatalError" that defaults to true.

### DIFF
--- a/lib/testLauncher/commands/automated.js
+++ b/lib/testLauncher/commands/automated.js
@@ -54,6 +54,11 @@ const builder = yargs => {
 		.option('log-level', {
 			describe: texts.cliLogLevel(),
 			global: false,
+		})
+		.option('die-on-fatal-error', {
+			describe: texts.cliDieOnFatalError(),
+			default: true,
+			global: false,
 		});
 };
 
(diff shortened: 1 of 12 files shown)


lib/testLauncher/commands/automated.js
```diff
@@ -54,6 +54,11 @@ const builder = yargs => {
 		.option('log-level', {
 			describe: texts.cliLogLevel(),
 			global: false,
+		})
+		.option('die-on-fatal-error', {
+			describe: texts.cliDieOnFatalError(),
+			default: true,
+			global: false,
```
andriichumak: Shouldn't there be `type: 'boolean',`?

lib/testLauncher/commands/interactive.js
```diff
@@ -59,6 +59,11 @@ const builder = yargs => {
 		.option('log-level', {
 			describe: texts.cliLogLevel(),
 			global: false,
+		})
+		.option('die-on-fatal-error', {
+			describe: texts.cliDieOnFatalError(),
+			default: true,
+			global: false,
```
andriichumak: Shouldn't there be `type: 'boolean',`?

andriichumak:
@givehug please include more info to PR title next time.

lib/testLauncher/commands/automated.js
givehug: updated

PR #80, opened 2018-06-21
SUIT-9076 - display progress messages with normal verbosity

andriichumak:
as manual action request in interactive mode is currently a progress message as well


andriichumak:
Yes, it's the reason I'm returning the old way of displaying progress messages back.

PR #82, opened 2018-06-25
SUIT-9746 remove --logDir from suitest help

AntonSBilan:
remove --logDir from suitest help


givehug:
I also noticed that '--help' and '--version' options are displayed as [boolean]

PR #85, opened 2018-06-26
SUIT-9725 Review TS definition with TS tests (demo set up and small fix of element chain)

AntonSBilan:
@andriichumak @givehug @JeKachan This PR is more like setup of UT for ts definitions, later more tests can be added, also maybe we should use https://github.com/Microsoft/dtslint


typeDefinition/constants/DomElementProps.d.ts
```diff
@@ -0,0 +1,75 @@
+export type DomElementProps = {
```
andriichumak: I would not name it `DomElementProps`, there not only DOM elements involved, but also xbox/tvOS/Android. Maybe just remove `Dom` part

typeDefinition/ElementChain.d.ts
```diff
@@ -13,8 +13,10 @@ import {
 	BaseChain,
 	UntilModifier,
 } from './modifiers';
+import {DomElementProps} from "./constants/DomElementProps";
 
 export interface ElementChain extends
+    DomElementProps,
```
andriichumak: Is it the correct place for it? It's what Element Query Promise returns, not a property of element chain.

testDefinition/examplesTS/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
+    "module": "system",
+    "maxNodeModuleJsDepth": 1000,
+    "preserveConstEnums": true,
+    "diagnostics": true,
+    "target": "es6",
+    "noImplicitAny": true,
+    "sourceMap": false,
```
givehug: source map is false by default

testDefinition/examplesTS/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "types": [
+      "../../index.d.ts"
+    ],
+    "noStrictGenericChecks": false,
```
givehug: noStrictGenericChecks  is false by default

typeDefinition/ElementChain.d.ts
JeKachan: @AntonSBilan I am not sure that this type should be here. Did you test it?

typeDefinition/ElementChain.d.ts
```diff
@@ -13,8 +13,10 @@ import {
 	BaseChain,
 	UntilModifier,
 } from './modifiers';
+import {DomElementProps} from "./constants/DomElementProps";
 
 export interface ElementChain extends
```
givehug: I think this will also allow to do `suitest.element().text` which is wrong. Should be probably for ElementEvalChain ? 
@JeKachan

JeKachan: @givehug I'am not sure. What should be returned for `await element('myButton').click()` ?

givehug:
Also these test should run together with all other ? or included in circle ci separately

typeDefinition/ElementChain.d.ts
andriichumak: `await ...click()` should resolve to `undefined` or maybe `true`.

givehug: yes ElementQueryChain, my mistake

andriichumak:
My idea was to simply include "typescript ./pathToTsFiles" to circle ci. Then if it returns wring exit code - pipe would fail. But using programmatic API is also OK. Doesn't matter as long as we receive compiler error message in console output

typeDefinition/ElementChain.d.ts
```diff
@@ -95,7 +96,8 @@ interface ElementEvalModifiers<T> extends
 {}
 
 interface ElementBaseQueryChain<TSelf> extends
-	BaseChain<TSelf, ElementQueryResult, ElementAbandonedChain>
+	BaseChain<TSelf, ElementQueryResult, ElementAbandonedChain>,
+	ElementProps
```
JeKachan: You should remove it. You need replace `type ElementQueryResult = string;` to `type ElementQueryResult = ElementProps;`

PR #87, opened 2018-06-28
Hotfix/suit 9759 Fix error stack trace

givehug:
Fix error stack trace
- wrapped chain promise callback with stackTraceWrapper to remember initial stack before executing async operations
- in makeChain do the same manipulation, create stack and then in processServerResponse prepend it to error 
Both are necessary because commands and chains are implemented differently. When/if we implement commands as chain, stackTraceWrapper can be dropped

### DIFF
--- a/lib/utils/makeChain.js
+++ b/lib/utils/makeChain.js
@@ -7,6 +7,8 @@ const makeChain = (getComposers, data) => {
 		const composers = getComposers(updatedData);
 		const chain = {};
 
+		updatedData.stack = Error().stack;
+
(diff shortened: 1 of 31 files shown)


lib/utils/makeChain.js
```diff
@@ -7,6 +7,8 @@ const makeChain = (getComposers, data) => {
 		const composers = getComposers(updatedData);
 		const chain = {};
 
+		updatedData.stack = Error().stack;
```
JeKachan: are `new` needed before `Error()` ?

givehug: new is not required

PR #93, opened 2018-07-06
SUIT-9856 Add illegal element selector validation to js api

JeKachan:
* add validation for addition properties
 * add proper error message for additional properties


JeKachan:
On my machine all tests passed. Also, I asked @givehug and @AntonSBilan run tests on their machines - test and coverage passed, as well.

PR #94, opened 2018-07-09
SUIT-9858 according to index.d.ts setAppConfig accepts only one argument

AntonSBilan:
according to index.d.ts setAppConfig accepts only one argument


index.d.ts
```diff
@@ -36,7 +36,7 @@ declare namespace suitest {
 		startTestPack(options: StartTestPackOptions): Promise<StartTestPackResult|SuitestError>;
 		openSession(options: OpenSessionOptions): Promise<OpenSessionResult|SuitestError>;
 		closeSession(): Promise<object|SuitestError>;
-		setAppConfig(options: SetAppContextOptions): Promise<void|SuitestError>;
+		setAppConfig(configId: string, options: SetAppContextOptions): Promise<void|SuitestError>;
```
andriichumak: You should probably as well edit `SetAppContextOptions`, it still has `configId` in it. Second parameter is simply config override.

PR #95, opened 2018-07-11
SUIT-9903 An error thrown by Suitest launcher when started with --log-dir option

AntonSBilan:
An error thrown by Suitest launcher when started with --log-dir option added code and path to error message


lib/utils/testLauncherHelper.js
```diff
@@ -124,13 +124,12 @@ function writeLogs(deviceId, userArgs, logDir, deviceName) {
 	try {
 		if (logDir) {
 			launcherLoggerHelper.mkDirByPathSync(logDir);
-			writeStream = launcherLoggerHelper.createWriteStream(logDir, deviceId, deviceName);
+			writeStream = launcherLoggerHelper.createWriteStream(logDir, deviceId);
```
andriichumak: So, this is the problem, stream is created from file twice? How did we miss this?

AntonSBilan: It solves problem on windows, my mac and Vladimirs ubuntu worked fine with it. Not sure how it could happened, I think it may be auto merge issue

andriichumak: Please, investigate. Also, since you're on the topic, I noticed that in console you'll receive 2 logs for each device. Once it will name the device, the second time same log with only device ID will be shown. Please, get rid of that second log, only leave the one with device name. Device ID is visible in log file path anyway.

PR #97, opened 2018-07-17
Hotfix/suit 9952 Improve error message for JS expression query chains.

givehug:
Fix error message for JS expression query chains.


andriichumak:
@givehug check out what CodeClimate is suggesting. Does it make sense to fix?

givehug:
@andriichumak I fixed 'too many return statements error', now it complains about 'Function deviceError has a Cognitive Complexity of 6 (exceeds 5 allowed).', should we fix these type of errors?

andriichumak:
Leave this one for now. We'll need to go through all the checks that CodeClimate does and ensure they make sense for our project. After that we'll start enforcing the rules in new PRs.

PR #98, opened 2018-07-18
SUIT-9969 Adjust JS API library to support Roku

JeKachan:


andriichumak:
Let's leave this for a while, there could also be some element properties specific to Roku.

PR #101, opened 2018-07-19
SUIT-9670 Improve networkRequest error description

givehug:
Improve networkRequest error description (actual / expected values)


andriichumak:
@givehug, see if it's possible to make CodeClimate happy.

PR #104, opened 2018-07-24
SUIT-10092 - Add CodeClimate badges to JS API repo

andriichumak:
After this merge we'll start to respect what CodeClimate suggest and will no longer merge PRs if CodeClimate maintainability decreases with your changes.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,35 @@
+version: "2"
+checks:
+  argument-count:
+    config:
+      threshold: 4
+  complex-logic:
+    config:
+      threshold: 4
+  file-lines:
+    config:
+      threshold: 250
+  method-complexity:
+    config:
+      threshold: 5
+  method-count:
+    config:
+      threshold: 20
+  method-lines:
+    config:
+      threshold: 25
+  nested-control-flow:
+    config:
+      threshold: 4
+  return-statements:
+    config:
+      threshold: 4
+  similar-code:
+    config:
+      enabled: false
+  identical-code:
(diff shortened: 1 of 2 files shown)


.codeclimate.yml
```diff
@@ -1,3 +1,35 @@
...
+checks:
+  argument-count:
+    config:
+      threshold: 4
+  complex-logic:
+    config:
+      threshold: 4
+  file-lines:
+    config:
+      threshold: 250
```
givehug: This will be quite challenging for some *.test.js files at least.

JeKachan:
I understand why needed similar-code, identical-code, but not understand why we need other restrictions (and where from they values where took) - not clear how they can improve code quality.

.codeclimate.yml
andriichumak: Well, this is a default value, so it was already like this for some time. I agree that for tests it's too small, but for production code I think this is a reasonable value. We could add tests into ignore list. I'll see what config options there are.

andriichumak:
@JeKachan those checks would help us not to be messy with our code. You can read about the concepts here and on adjusting pages:
https://docs.codeclimate.com/docs/cognitive-complexity

andriichumak:
Values for those restrictions are based on default ones. If we will feel that some values are too strict or too loose - we'll adjust the config. For now it seams reasonable.

andriichumak:
So, from now on we will not merge unless CodeClimate says it's OK. If some rules will start to bother you and you think they are unfair - we'll review those.

PR #105, opened 2018-07-24
SUIT-9013 - Add CONTRIBUTING.md

andriichumak:


CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
+# Contributing
+
+First off, thank you for considering contributing to Suitest JS API.
+
+When contributing to this repository, please first discuss the change you wish to make via GitHub issue
+with the owners of the repository.
+
+Please note we have a [code of conduct][]. Follow it in all your interactions with the project.
```
SuitLollipop: where is that code? I think that at this stage it may be unnecessary.

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 1. How do I start contributing?
+
+If you've noticed a bug, have a question or a suggestion, [search the issue tracker][] to see
```
SuitLollipop: would this link work?

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+if someone has already create a ticket. If not, go ahead and [make one][new issue].
+
+## 2. Fork & create a branch
+
+If this is something you think you can fix, then [fork Suitest JS API][] and
```
SuitLollipop: what about this?

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+create a branch with a descriptive name.
+
+We use `hotfix/[issue-id]-[issue-title]` convention for small changes. For example:
```
SuitLollipop: i don't think we need to force this convention onto the contributers, especially since they don't have access to THE issue tracker.

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+``` 
+
+## 3. Work on the issue
+
+Check out your fork, switch to the newly created branch and start fixing the issue.
+
+Here are the steps to get you started:
+
+* Make sure you have [NodeJS][] of v10 or higher installed.
+* Make sure you have [Yarn][] installed globally. You can use NPM instead, if you'd like.
```
SuitLollipop: links

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+* Make sure to run `yarn lint` (or `npm run lint`) before making commit.
+
+> **Heads up!** In order to make unit tests pass, you should remove (or temporary rename)
+> all `.suitestrc` files starting from project's dir and up the folder structure, as
+> well as in your home folder. We will updated our unit tests to account for that,
+> but for now you'll have to use this workaround.
+
+## 4. Prepare the pull request
+
+After you're done coding, commit and push your changes to your fork. In GitHub, [create a pull request][]
```
SuitLollipop: link

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+  code maintainability must not decrease after pull request merge.
+* At least 2 members of Suitest Core team approved the pull request.
+
+[code of conduct]: https://github.com/SuitestAutomation/suitest-js-api/blob/master/CODE_OF_CONDUCT.md
+[search the issue tracker]: https://github.com/SuitestAutomation/suitest-js-api/issues?q=something
+[new issue]: https://github.com/SuitestAutomation/suitest-js-api/new
+[fork Suitest JS API]: https://help.github.com/articles/fork-a-repo
+[NodeJS]: https://nodejs.org/
+[Yarn]: https://yarnpkg.com/
+[create a pull request]: https://help.github.com/articles/creating-a-pull-request
```
SuitLollipop: i see that my comments about links are invalid. If only these would have been at the top of the file :)

CONTRIBUTING.md
andriichumak: There is a link to it. It's in the bottom. This is the way GitHub links are usually made - you have [text][] in body and then [text]: http://... in footer.

CONTRIBUTING.md
andriichumak: yes

CONTRIBUTING.md
andriichumak: By issue tracker I mean the GitHub one's. The only part that makes sense to drop is "hotfix/", but that's a not big deal anyway, right?

andriichumak:
@SuitLollipop I would prefer to still have "hotfix/" part in branch name, so that we have same convention everywhere. I don't think contributor would mind typing a few extra keys. And we can be less restrictive on this matter when dealing with external contributors.

PR #106, opened 2018-07-27
SUIT-10151 Fix element chain typescript definition

JeKachan:


typeDefinition/ElementChain.d.ts
```diff
@@ -107,4 +107,4 @@ interface ElementBaseEvalChain<TSelf> extends
 interface ElementAbandonedChain extends AbstractChain {}
 
 type ElementQueryResult = ElementProps;
-type ElementEvalResult = void;
+type ElementEvalResult = boolean | void;
```
givehug: Lets do it for all chains?

givehug:
@andriichumak @JeKachan 
- Please check my last two commits. 
- Test coverage failed on all of the recent commits. Same on roku branch. It very rarely fails on by machine. Mostly somewhere in SuitestLauncher.test.js. We should consider refactoring those tests, stubbing all logs and nocking everything.

andriichumak:
@givehug please, do the refactoring as suggested. We were postponing it for a long time. That should probably be a new hotfix.

andriichumak:
@givehug typescript checks are failing now. Please, fix.

givehug:
Updated

JeKachan:
Approve

typeDefinition/PositionChain.d.ts
```diff
@@ -46,4 +46,4 @@ interface PositionEmptyChain extends PositionBaseEvalChain<PositionEmptyChain> {
 interface PositionAbandonedChain extends AbstractChain {}
 
 type PositionQueryResult = string;
```
AntonSBilan: is it correct? for me 
const el = await suitest.position(10, 10);
		console.log(typeof el); // prints `boolean`

givehug: I think, there is no query for position, only cookie, element, location and jsExpression.

PR #112, opened 2018-08-10
hotfix/suit 9394 Improve unit tests stability

givehug:
Added spawn mocking
Cleared console output from redundant logs
Improved launcher tests stability


test/api/webSockets.test.js
```diff
@@ -85,6 +87,7 @@ describe('webSockets', () => {
 
 		assert.ok(err);
 		assert.strictEqual(webSockets.isConnected(), false, 'connected');
+		console.error.restore();
```
AntonSBilan: If test fails during execution this line may not be called, probably better do this in after or afterEach

PR #113, opened 2018-08-10
SUIT-10291 Handle test line warnings

givehug:
Treat test line result with status warning as non error.

### DIFF
--- a/lib/utils/socketChainHelper.js
+++ b/lib/utils/socketChainHelper.js
@@ -22,6 +22,10 @@ const processServerResponse = toString => function processServerResponseHandler(
 		toString,
 	});
 
+	// warnings
+	if (res.result === 'warning') {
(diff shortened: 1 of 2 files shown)


lib/utils/socketChainHelper.js
```diff
@@ -44,19 +48,19 @@ const processServerResponse = toString => function processServerResponseHandler(
 		}
 	}
 	// eval success
-	if (res.result === 'success' && res.contentType === 'eval') {
+	if (isSuccess(res.result) && res.contentType === 'eval') {
```
JeKachan: all nice for me, but i think that `isSuccess` is redundant:
as `isSuccess` used only inside `processServerResponseHandler` we can drop this function and create variable instead:
`const isSuccess = res.result === 'success' || res.result === 'warning';`
als we can create the similar const for eval line:
`const isEval = res.contentType === 'eval'` e.t.c

PR #114, opened 2018-08-14
suit 10210 drop symbols and symbol mapping

AntonSBilan:
suit 10210 drop symbols and symbol mapping


lib/validataion/jsonSchemas.js
```diff
@@ -365,6 +379,15 @@ schemas[validationKeys.ENV_VARS_INTERACTIVE] = {
 	},
 };
 
+schemas[validationKeys.ARRAY_OF_BUTTONS] = {
```
givehug: Could be generalized to ARRAY_OF_STRING or non empty strings

AntonSBilan: Added button validation

lib/chains/networkRequestChain.js
```diff
@@ -96,21 +89,21 @@ const toJSON = data => {
...
-	subject.compare = SUBJ_COMPARATOR_M[data.comparator.type];
+	subject.compare = data.comparator.type;
 	subject.val = data.comparator.val;
 
 	// Apply request parameters
 	subject.requestInfo = (data.request && data.request.props || []).map(item => ({
-		name: typeof item.name === 'symbol' ? NETWORK_PROP_M[item.name] : item.name,
-		val: item.name === NETWORK_PROP.METHOD ? NETWORK_METHOD[item.val] : item.val,
-		compare: PROP_COMPARATOR_M[item.compare],
+		name: item.name,
```
andriichumak: Do we actually need to map in this case? Are there any other properties in item?

lib/chains/networkRequestChain.js
```diff
@@ -96,21 +89,21 @@ const toJSON = data => {
...
+		compare: item.compare,
 	}));
 
 	// Apply response parameters
 	subject.responseInfo = (data.response && data.response.props || []).map(item => ({
-		name: typeof item.name === 'symbol' ? NETWORK_PROP_M[item.name] : item.name,
+		name: item.name,
 		val: item.val,
-		compare: PROP_COMPARATOR_M[item.compare],
+		compare: item.compare,
```
andriichumak: Is mapping required here?

lib/constants/element.js
```diff
@@ -1,94 +1,105 @@
...
+	CONTENT_MODE: ELEMENT_PROP.CONTENT_MODE,
+	TEXT_ALIGNMENT: ELEMENT_PROP.TEXT_ALIGNMENT,
 };
 
 Object.freeze(ELEMENT_PROP);
+Object.freeze(VALUES);
 
 // Element special values
-const VALUE = {REPO: Symbol('valueRepo')};
+const VALUE = {REPO: 'valueRepo'};
```
andriichumak: This one is an exception and should stay Symbol. Because user might actually want to use "valueSymbol" as a value.

lib/constants/networkRequest.js
```diff
@@ -1,21 +1,23 @@
+/* eslint-disable */
+
 const NETWORK_PROP = {
-	BODY: Symbol('body'),
-	METHOD: Symbol('method'),
-	STATUS: Symbol('status'),
+	BODY:   '@body',
```
andriichumak: This should stay symbols as well. "@" is not a legal char for header, however I can imaging non-standard implementations where it's allowed.

test/utils/socketErrorMessages.test.js
```diff
@@ -89,86 +89,86 @@ describe('Socket error messages', () => {
...
-						type: 'networkRequest',
-						comparator: {val: 'test'},
-					},
-				},
-				'queryFailed: "Chain description."'
-					+ `${EOL}\tFailing checks:`
-					+ `${EOL}\t~ url: test (expected)`
-					+ `${EOL}\t× url: request was not made (actual)`,
-			],
+			// [basePayload('failedStart'), 'Chain description.'],
```
JeKachan: Why here commented?

lib/constants/videoState.js
```diff
@@ -1,17 +1,19 @@
...
+	ERROR:              'error',
+	PLAYING:            'playing',
+	PAUSED:             'paused',
+	CONNECTING:         'connecting',
+	BUFFERING:          'buffering',
+	FINISHED:           'finished',
+	IDLE:               'idle',
+	PREPARING:          'preparing',
+	PREPARED:           'prepared',
+	PLAYBACK_COMPLETED: 'playback_completed',
```
andriichumak: Please, use camelCase everywhere.

AntonSBilan: Do you mean change properties names like `playbackCompleted: 'playback_completed'` and  `prepared: 'prepared'`?, it will brake existing tests

andriichumak: It should be:
PLAYBACK_COMPLETED: 'playbackCompleted'

AntonSBilan: Should vrc buttons and network methods be in camelCase as well?

andriichumak: Nope, those are all uppercase. all constants should be either camelCase or UPPER_CASE_WITH_DASH, but never lower_case_with_dash

AntonSBilan:
@andriichumak 'playback_completed' was only one prop with inappropriate  name format, do we already have a ticket for BE to rename it?

lib/constants/videoState.js
```diff
@@ -1,17 +1,19 @@
...
+	ERROR:              'error',
+	PLAYING:            'playing',
+	PAUSED:             'paused',
+	CONNECTING:         'connecting',
+	BUFFERING:          'buffering',
+	FINISHED:           'finished',
+	IDLE:               'idle',
+	PREPARING:          'preparing',
+	PREPARED:           'prepared',
+	PLAYBACK_COMPLETED: 'playbackCompleted',
```
andriichumak: @AntonSBilan I talked to BE, they are not willing to make comparison case-insensitive for religious reasons. Please, revert this to "playback_completed"

andriichumak:
@AntonSBilan did you resolve what's in Ievgen's comments?

AntonSBilan:
@andriichumak yes, all comments should be resolved

PR #115, opened 2018-08-14
adding timestamps

jan-tomasek-iflix:
Feature adds timestamp in a format YYYYMMDDTHHmmss to the log line the new message is in format:

```<device_name> <timestamp> <message>```

### CODE REVIEW

**Comment by andriichumak** [comment]
Hello Jan. Thanks for the pull request. We'll look into this asap.

**Comment by givehug** [comment]
How about making this optional with some new config field?

**Comment by jan-tomasek-iflix** [comment]
You mean adding the parameter to the commandline binary (bin/suitest) that would customise the format? Enable and disable some custom log line format? Or do you mean some log line formatter something like winston?

File: lib/utils/logger.js [hunk 1]
```diff
@@ -119,7 +122,7 @@ const log = method => (message, deviceId) => {
 			: stColors.suit(devicePrefix(deviceId)) + prefix + message
```
andriichumak: @jan-tomasek-iflix could you please move `currentTimestamp` to lines 121 and 122, see that it's prepended before "Device XXXXX>" string

lib/utils/logger.js
```diff
@@ -93,6 +94,8 @@ Object.freeze(methods);
  * @param {*} method
  */
 const log = method => (message, deviceId) => {
+	const currentTimestamp = timestamp('YYYYMMDDTHHmmss');
```
andriichumak: @jan-tomasek-iflix could you please update timestamp format to `YYYY.MM.DD HH:mm:ss` and wrap it to square brackets? So that resulting string would be something like `[2018.08.14 12:40:53]`.
One of our devs will make format configurable through CLI params and RC file after this pull request is merged, so you'll still be able to use timestamp in your preferred format.

andriichumak:
@jan-tomasek-iflix Could you please drop `package-lock.json` file from this PR? We are using Yarn and it's enough to have `yarn.lock`.
I turned ON our CircleCI integration for pull requests, it was off for some reason... So, with your next commit it should also run unit tests automatically.

andriichumak:
@givehug We will make timestamp configurable with another pull request.

andriichumak:
@jan-tomasek-iflix thanks, merged. We'll release this today or tomorrow.

PR #116, opened 2018-08-15
SUIT-10354 Make timestamp in logs configurable with cli or rc file

givehug:
Made timestamp configurable with CLI args or RC file

### DIFF
--- a/lib/testLauncher/commands/common.js
+++ b/lib/testLauncher/commands/common.js
@@ -1,0 +1,39 @@
+const texts = require('../../texts');
+const logLevels = require('../../constants/logLevels');
+
+/**
+ * Apply cli args common for interactive and automated commands
+ * @param {*} yargs
+ */
+function applyCommonArgs(yargs) {
+	yargs
+		.option('continue-on-fatal-error', {
+			describe: texts.cliContinueOnFatalError(),
+			default: false,
+			global: false,
+			type: 'boolean',
+		})
+		.option('disallow-crash-reports', {
+			describe: texts.cliDisallowCrashReports(),
+			global: false,
+			type: 'boolean',
+		})
+		.option('log-dir', {
+			describe: texts['tl.logDirDescription'](),
+			global: false,
+		})
+		.option('log-level', {
+			describe: texts.cliLogLevel(),
+			global: false,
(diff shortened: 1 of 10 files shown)


lib/utils/logger.js
```diff
@@ -94,7 +94,7 @@ Object.freeze(methods);
  * @param {*} method
  */
 const log = method => (message, deviceId) => {
-	const currentTimestamp = `[${timestamp('YYYY.MM.DD HH:mm:ss')}]`;
+	const currentTimestamp = config.timestamp === 'none' ? '' : `[${timestamp(config.timestamp)}] `;
```
andriichumak: Please, move "none" to constant and make comparison case-insensitive

lib/texts.js
```diff
@@ -251,6 +251,7 @@ ${leaves}`,
 	cliLogLevel: () => 'Logger verbosity level',
 	cliDisallowCrashReports: () => 'Disallow Suitest to submit automatic crash reports when unexpected exceptions occur',
 	cliContinueOnFatalError: () => 'Do not force process exit when suitest chain fails with fatal error',
+	cliTimestamp: () => 'Logger timestamp format. Use "none" to disable',
```
andriichumak: Some example or at least default value would be nice. Otherwise user wouldn't have a clue on how to define format.

lib/testLauncher/commands/common.js
```diff
@@ -0,0 +1,39 @@
+const texts = require('../../texts');
```
andriichumak: Good idea to move that to `common.js`

lib/texts.js
andriichumak: Not sure if `yargs` displays default value in some form. Perhaps it's enough to put default value there.

givehug: Yargs supports default option, but unfortunately with current implementation we override config with cli args, so if user has something in RC file it will be overridden by default option if the same cli arg is not present. Similar to logLevel, we don't specify default option for it.

andriichumak: OK, then default value should be in description

givehug:
Updated, also fixed issue of overriding config with undefined values from cli or rc file

PR #120, opened 2018-08-21
Hotfix/suit 10382 Add interactive REPL for debugging

givehug:
Added .interactive() command, which will pause test execution and open interactive REPL for better debugging.
When in repl 'suitest.resume()', will exit repl and continue test execution.
Command is allowed in interactive session mode only with --repl flag.
Timestamp + deviceId prefixes wont be displayed in interactive session with --repl flag, similar to how it is in debug mode.

Note: you have to use npm script similar to debug in this case like "suitest interactive --repl node ./node_modules/mocha/bin/_mocha --no-timeouts --exit". I haven't found a way to make it work with simply 'mocha'.


test/utils/testLauncherHelper.test.js
```diff
@@ -57,26 +57,6 @@ describe('testLauncherHelper util', () => {
 		assert.deepStrictEqual(testLauncherHelper.mergeConfigs({a: true}, {a: false}), {a: false});
 	});
```
givehug: Removed this test because it started to fail very often for me. I is refactored in release sp 2 already.

lib/testLauncher/repl.js
```diff
@@ -0,0 +1,38 @@
...
+function startRepl(suitest) {
+	const replInstance = repl.start({
+		prompt: '> ',
+	});
+
+	isActive = true;
+	replInstance.context.suitest = suitest;
+
+	return new Promise(resolve => {
+		suitest.resume = () => {
```
JeKachan: `resume` method available only after calling `suitest.interactive()`?

givehug: yes, only in repl

givehug: I just realized, user is also able to exit repl with `.exit` command, 
https://nodejs.org/api/repl.html#repl_commands_and_special_keys
I'll investigate, suitest.resume could be obsolete.

lib/testLauncher/SuitestLauncher.js
```diff
@@ -180,6 +181,15 @@ class SuitestLauncher {
 							execPath: normalize(process.execPath),
 						}).on('close', resolve);
 
+						registerProcess(child);
+					} else if (repl) {
```
AntonSBilan: @givehug so we can be in debug or in repl is there a validation for that. It may be confusing that presents of `inspect` in suitest args prevent repl mode

givehug: There is no validation now. Yes it can be one of two only. At least in VS code repl can't be opened in debug console.

AntonSBilan: @givehug do you think it can be useful to have debug and repl mode at same test run. The is also a chrome devtools which should not block repl

andriichumak: It could be useful. We should test it with Chrome. If it's not possible to combine debug and repo modes there as well - we should warn user about it and close the process.

lib/testLauncher/commands/interactive.js
```diff
@@ -48,6 +48,12 @@ const builder = yargs => {
 		.option('inspect-brk', {
 			describe: 'Will launch user command with --inspect-brk execArgv, used for debugging',
 			global: false,
+		})
+		.option('repl', {
+			describe: 'Will allow interactive REPL, used for debugging',
```
andriichumak: ... used for test authoring

lib/testLauncher/repl.js
andriichumak: If `.exit` would only stop REPL and normal execution would resume - sure. But if it's exiting the process - we need `suitest.resume()` still.
There is an option to add your own syntax for special commands I think. Would also be an option.

andriichumak:
@givehug We should update `yarn.lock` only when something changes in `package.json`.

lib/testLauncher/SuitestLauncher.js
givehug: yes it is possible

lib/testLauncher/repl.js
givehug: .exit doesn't close process

andriichumak:
@givehug please, resolve conflicts.

honeydripper:
@givehug @andriichumak let's wait with any efforts on this. During my experiments the original specs have changed a lot and furthermore, I have a version of REPL already working with the new logs. Will keep you posted.

PR #121, opened 2018-08-22
SUIT-10399 Errors thrown by calling SuitestError exit method displays without colors

AntonSBilan:
SUIT-10399 Errors thrown by calling SuitestError exit method displays without colors


andriichumak:
@AntonSBilan there is no reason for `yarn.lock` to be committed here, as there were no changes to `package.json`. Could you please drop it?

PR #125, opened 2018-09-04
SUIT-10518 TypeError: Cannot read property 'type' of undefined

AntonSBilan:
TypeError: Cannot read property 'type' of undefined


lib/texts.js
```diff
@@ -151,6 +151,7 @@ ${leaves}`,
 	'errorType.syntaxError.WrongDelay': template`Test command received invalid input. Delay value is invalid. ${0}`,
 	'errorType.syntaxError.wrongBody': template`Body field value is exceeding 4KB limit. ${0}`,
 	'errorType.syntaxError.wrongUrl': template`This does not look like a valid URL. ${0}`,
+	'errorType.syntaxError.comparatorMissing': template`${0} chain requires comparator.`,
```
JeKachan: "Comparator" term are understandable for us, but not for user. Lets use the same terms as in docs: https://suite.st/docs/suitest-api/chain-modifiers/

JeKachan: I thinking from position google searching. Try to type in google: "suitest chain modifiers" and you will get proper page, but "suitest comparator" not work as expected. So, my purpose to use something like: "$0 chain modifiers are missing"

PR #127, opened 2018-09-05
SUIT-10511 Add new Roku properties

AntonSBilan:
Add new Roku properties


AntonSBilan:
@andriichumak Locally all tests pass successfully, probably cache

PR #131, opened 2018-09-28
SUIT-10538 Show application logs in JS API output

JeKachan:


lib/api/webSockets.js
```diff
@@ -54,6 +57,15 @@ function connect(connectionOps = {}) {
 			);
 		}
 
+		if (content.type === 'console' && content.data) {
+			const method = {
+				info: 'info',
+				log: 'log',
+			}[content.subtype] || 'debug';
+
+			return logger[method](content.data.join('\n'));
```
honeydripper: We need to mark the app logs somehow so that they are easily distinguishable from the API logs. 
We now use |A| for marking assertions and |E| for evaluations, Perhaps we should use |L| for for app logs (maybe also with a different color)

lib/api/webSockets.js
```diff
@@ -35,10 +36,12 @@ function connect(connectionOps = {}) {
 			// ignore
 		}
 
+		const content = path(['content'], obj) || {};
```
honeydripper: will this work if obj is undefined?

lib/api/webSockets.js
```diff
@@ -54,6 +57,15 @@ function connect(connectionOps = {}) {
 			);
 		}
 
+		if (content.type === 'console' && content.data) {
+			const method = {
+				info: 'info',
+				log: 'log',
+			}[content.subtype] || 'debug';
+
+			return logger[method](`|L| ${method}:`, content.data.join('\n'));
```
honeydripper: now that we introduce more type markup, I think it should have been placed in constants somewhere, before we make too much mess. Plus, please see the feedback in JIRA.

lib/api/webSockets.js
```diff
@@ -58,12 +58,19 @@ function connect(connectionOps = {}) {
...
 
 		if (content.type === 'console' && content.data) {
-			const method = {
-				info: 'info',
-				log: 'log',
-			}[content.subtype] || 'debug';
-
-			return logger[method](`|L| ${method}:`, content.data.join('\n'));
+			const logType = {
+				info: '|AI|',
```
honeydripper: these should have been constants (probably in opType.js) along with other used markup.

lib/api/webSockets.js
```diff
@@ -58,12 +58,19 @@ function connect(connectionOps = {}) {
...
+				log: '|AL|',
+				warn: '|AW|',
+				error: '|AE|',
+			}[content.subtype] || '|AL|';
+
+			// ADDING logType after new line characters needed for proper displaying multilines
+			// TODO: need to remove replacing after release test-repl-with-updated-logs branch
+			return logger.log(
+				logType,
+				content.data.join('\n').replace(/\n/g, `\n${logType} `)
```
honeydripper: if `content.data` is an array, you could have written it just as `...content.data`

lib/utils/logger.js
```diff
@@ -183,7 +185,15 @@ function outputLogsInLauncherMode(method, messages) {
 	}
 
 	const timeAndDevice = timestamp() + ' ' + logSource;
-	const leftRail = stColors.gray(timeAndDevice) + stColors.bold(opType);
+
+	const opTypeColor = {
+		'AL': stColors.bold,
```
honeydripper: these should have been used as constants (see previous comment)

lib/utils/opType.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
 	 * @returns {Boolean|null}
 	 */
 	isOpType: function(logString) {
-		return logString.match(/^\|[AEL]\|$/);
+		return logString.match(/^\|[AE]|L|AL|AW|AE|AI\|$/);
```
honeydripper: |L| should never happen, and again these could have been constants.

lib/api/webSockets.js
JeKachan: and only first line of multi-line log will be properly marked. We decide yesterday that we need insert markers after each new lines before branch test-repl-with-updated-logs will be release.

lib/utils/opType.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
 	 */
 	formatOpType: function(logString) {
 		const str = stripAnsiChars(logString);
-		const withOpType = str.match(/\|([AE])\|(.*)/);
+		const withOpType = str.match(/\|([AE]|AL|AW|AE|AI)\|(.*)/);
```
honeydripper: these should have used constants

lib/utils/opType.js
```diff
@@ -47,4 +47,12 @@ module.exports = {
 			logString: withOpType[2],
 		};
 	},
+	opTypes: {
+		assertLine: 'A',
```
honeydripper: you have introduced this constant but it is not used anywhere

lib/api/webSockets.js
honeydripper: i thought it did not work with multiline string but *did* work with multiple arguments, but I guess it does not.
Anyway then `content.data.forEach(one => logger.log(logType, one))` could still be a better option to string replacing.

JeKachan: I believe that this is not crucial point, anyway it will be changed after repl release.

lib/utils/opType.js
JeKachan: I made it explicitly for other developers for presenting which operation types currently used in library.

lib/api/webSockets.js
givehug: there is arrToString helper somewhere

lib/utils/opType.js
honeydripper: A and E are already used somewhere, I would expect this PR to refactor those usages into usages of this introduced constant.

lib/api/webSockets.js
honeydripper: I don't mind if we leave that, repl is in the late stage of its backing anyway

lib/utils/opType.js
```diff
@@ -7,34 +7,47 @@ const stripAnsiChars = require('./stringUtils').stripAnsiChars;
...
+const opTypes = Object.freeze({
+	assertLine: 'A',
+	evalLine: 'E',
+	appLog: 'AL',
+	appInfo: 'AI',
+	appWarn: 'AW',
+	appError: 'AE',
+});
+
+const baseOpTypeRegexSentence = '[AE]|AL|AW|AE|AI';
```
honeydripper: I would expect here Object.keys(opTypes).join('|'), but ok

PR #134, opened 2018-10-01
Release/sp 2 andrii

andriichumak:


AntonSBilan:
@givehug it seems correct and working with strings, but we may want to test it before release to make sure all error messages are correct

lib/chains/elementChain.js
```diff
@@ -89,7 +93,7 @@ const toString = (data, nameOnly = false) => {
 				break;
 			default: {
 				const matches = data.comparator.props.map(
-					one => `  ${getAllSymbols[one.name]} ${getAllSymbols[one.type]} ${getAllSymbols[one.val] || one.val}`
+					one => `  ${one.name} ${one.type} ${VALUE_M[one.val] || one.val}`
```
honeydripper: This does not look correct (unless overwritten in later commits). The other line should have been taken.

givehug: there are no symbols anymore, only VALUE

honeydripper: right.
@AntonSBilan what about validationKeys.js? 

Not sure if some of those could appear in some of the output logs. If yes they will need to be translated.

andriichumak:
@givehug, could you please resolve conflicts? We should have released this changes already.

lib/chains/pressButtonChain.js
```diff
@@ -46,6 +44,7 @@ const toString = data => {
 	return out;
 };
 const beforeSendMsg = (data) => logger.log(getOpType(data), toString(data));
+// const beforeSendMsg = (data) => logger.info(chainIsAboutToBeExecuted(`pressButton(${data.ids.join(', ')})`, toString(data)));
```
andriichumak: Drop the commented line
```suggestion
```

lib/constants/borderStyle.js
```diff
@@ -1,21 +1,23 @@
+/* eslint-disable */
```
andriichumak: We should not disable eslint completely, if necessary - disable specific rules. Same for other files below.

andriichumak: @givehug could you please as well fix this.

givehug:
@andriichumak  merged

PR #135, opened 2018-10-03
SUIT-10382 - implement REPL in interactive mode

honeydripper:
- Implemented `suitest.interactive()` command for entering a REPL shell 
- Got rid of `fork` now using exclusively `spawn`
- Moved device details to `pairedDeviceContext`
- Simplified `logger`


lib/testLauncher/repl.js
```diff
@@ -0,0 +1,174 @@
...
+		);
+	});
+
+	return new Promise(resolve => {
+		ipcSocket.on('data', chunk => {
+			const msg = chunk.toString().split('|');
+			const [ttyFlag, ttyVal] = msg;
+
+			if (ttyFlag === REPL_INIT)
+				return resolve(setupRepl(opts, !!+ttyVal, ipcSocket));
```
givehug: It resolves with another Promsie, but ipcSocket will keep listening for 'data' ?

lib/testLauncher/repl.js
```diff
@@ -0,0 +1,174 @@
...
+
+	if (!process.env.REPL_IPC_PORT) {
+		throw new SuitestError(
+			replIpcNotAvailable(),
+			SuitestError.IPC_ERROR
+		);
+	}
+
+	const ipcSocket = net.connect(process.env.REPL_IPC_PORT, 'localhost').unref();
+
```
givehug: We set up the ipcSocket  to make it work with spawn ?

lib/utils/chainPromise.js
```diff
@@ -18,7 +17,7 @@ let currentPromise = Promise.resolve();
  */
 
 const chainPromise = callback => (...args) => {
-	const wrappedCallback = stackTraceWrapper(() => callback(...args));
```
givehug: this was here so that commands (those in commands/ dir) have correct error stack, does it work without it ?

lib/utils/testLauncherHelper.js
```diff
@@ -243,17 +187,14 @@ async function launchChild(cmdArgv, ownArgv, options, runMode) {
...
+		args.unshift(`--inspect${inspectBrk ? '-brk' : ''}=${debugPort}`);
+	}
+
+	if (runMode === INTERACTIVE) {
+		args.unshift('--experimental-repl-await');
+		options.stdio = ['pipe', 'pipe', 'pipe'];
+		options.env.REPL_IPC_PORT = await setupReplIpc();
 	}
 
 	return cp.spawn(
```
givehug: Wasn't able to run repl interactive mode and debug in vscode, see ticket.

package.json
```diff
@@ -50,21 +50,24 @@
   },
   "typings": "index.d.ts",
   "devDependencies": {
+    "bluebird": "^3.5.2",
```
givehug: I only found the use of it in tests, and I think native Promise should work there, is it really needed?

lib/testLauncher/repl.js
honeydripper: Not sure what you mean with "another promise" but yes, the socket will remain alive

lib/testLauncher/repl.js
honeydripper: We need a channel between the main process and the child in order for the main process to know when it should switch the terminal to the raw mode and back. Mode switching is necessary so that the control commands go directly to the child process and NOT go to the main process.

When we were using fork node was managing this for us, however using fork with shared io does not work as expected and with piped io it kind of did although the terminal behaved weird from time to time. Plus the inconvenience with not being able to use the nonjs command launchers such as "mocha".

IPC communication looks like the following:
- Main process sets up TCP server
- Main process passes the server port number to the child process via env variable.
- Child process connects (when suitest.interactive starts) and submits a handshake
- Main process responds with the isTTY value (since it is not available for reading from the child process)
- Child process sets up REPL terminal and sends message again
- Main process switches terminal to raw mode.
- On repl exit child process sends another notification and main process repipes io and switches raw mode off.

As a bonus if ever the API should be ported to another language the communication can stay the same.

lib/commands/interactive.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {{ignored: *, watch: *, repeater: Function? cwd: String?, vars: Array?}} opts
+ * @returns {*[]}
+ */
+function getReplParameters(opts) {
+	// get dir where .interactive() was called
+	const callFile = path.normalize(stackTraceParser(new Error())[2].file);
+	const cwd = path.dirname(callFile);
+	const files = opts.watch || path.join(cwd, '**/*.js');
+	const watch = Array.isArray(files)? files : [files? files : cwd];
+	const suitest = this;
```
JeKachan: As I noticed `getReplParameters` is used only in `interactive` function. So `this` will be refers to `global` object? why just not do this explicitly?

lib/commands/interactive.js
```diff
@@ -0,0 +1,70 @@
...
+	const vars = Object.assign({suitest}, opts.vars);
+
+	return [vars, cwd, watch];
+}
+
+/**
+ * Start interactive REPL
+ * @param {{ignored: *, watch: *, repeater: Function? cwd: String?, vars: Array?}} opts
```
JeKachan: `ignored` and `watch` has unknown type? Comma needed before `cwd`

lib/commands/interactive.js
```diff
@@ -0,0 +1,70 @@
...
+async function interactive(opts = {}) {
+	if (isReplActive)
+		return Promise.resolve();
+
+	const {replWarnInteractive, replWelcomeMessage, replSessionEnded} = texts;
+
+	if (!authContext.isInteractiveSession()) {
+		logger.warn(replWarnInteractive());
+
+		return Promise.resolve();
```
JeKachan: No need to return Promise.resolve() because function already async and all returned values will be wrapped in Promise.

lib/utils/logger.js
```diff
@@ -92,13 +98,14 @@ Object.freeze(methods);
 
 /**
  * Prefix log with device id
- * @param {object} deviceDetails
  */
-function devicePrefix(deviceDetails) {
-	if (!deviceDetails || !deviceDetails.shortDisplayName)
+function devicePrefix() {
```
JeKachan: I dont like  switching from pure function to dirty. This function now will be hard to test and debugg, because now it completely depends on environment presetup.

lib/utils/socketChainHelper.js
```diff
@@ -21,18 +20,27 @@ const processServerResponse = toString => function processServerResponseHandler(
...
 	});
+	const throwErr = err => {
+		err.stack = data.stack;
+		throw err;
+	};
+	const infoMessage = prefix => getInfoErrorMessage(message, prefix, res, data.stack);
+	const isSuccess = res.result === 'success' || res.result === 'warning';
+	const isEval = res.contentType === 'eval';
+	const isTestLine = res.contentType === 'testLine';
+	const isQuery = res.contentType === 'query';
```
JeKachan: By these examples you also can create `isWarning` and `isFatal` variables

lib/utils/testHelpers/repl.js
```diff
@@ -0,0 +1,2 @@
+// asset file for testing repl
```
JeKachan: this file still needed?

lib/utils/testLauncherHelper.js
```diff
@@ -248,17 +191,14 @@ async function launchChild(cmdArgv, ownArgv, options, runMode) {
...
-			execPath: normalize(process.execPath),
-			silent: true,
-		});
+		args.unshift('--inspect-brk');
+		args.unshift(`--inspect${inspectBrk ? '-brk' : ''}=${debugPort}`);
+	}
+
+	if (runMode === INTERACTIVE) {
+		args.unshift('--experimental-repl-await');
+		options.stdio = ['pipe', 'pipe', 'pipe'];
```
JeKachan: from nodejs docs:
'pipe' - equivalent to ['pipe', 'pipe', 'pipe'] (the default)

lib/utils/testLauncherHelper.js
```diff
@@ -268,6 +208,114 @@ async function launchChild(cmdArgv, ownArgv, options, runMode) {
 	);
 }
 
+/**
+ * Save child output to the file
+ * @param {ChildProcess} child
+ * @param {Buffer} chunk
+ * @param {WritableStream?} logStream
```
JeKachan: what is the point of making logStream optional if without it function will not work? I think it should be marked as required for proper IDE support

test/commands/interactive.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+		return new Promise(async resolve => {
+			startRepl = sinon.stub(repl, 'startRepl').callsFake(async() => {
+				await interactive();
+				++replCalls;
+				assert.equal(replCalls, 1, 'Second call is ignored');
+				resolve();
+			});
+			await interactive();
+		}).then(async() => {
```
JeKachan: Why just not await promise above without additional chains or mocha will not close process?

test/testLauncher/repl.test.js
```diff
@@ -0,0 +1,139 @@
...
+
+		delete process.env.REPL_IPC_PORT;
+
+		const {startRepl} = suitestRepl;
+		let passed = false;
+
+		try {
+			startRepl();
+		} catch (e) {
+			console.log(replIpcNotAvailable.called);
```
JeKachan: this console used for debug?

test/testLauncher/repl.test.js
```diff
@@ -0,0 +1,139 @@
...
+		const replIpcNotAvailable = sandbox.stub(texts, 'replIpcNotAvailable');
+		const port = process.env.REPL_IPC_PORT;
+
+		delete process.env.REPL_IPC_PORT;
+
+		const {startRepl} = suitestRepl;
+		let passed = false;
+
+		try {
+			startRepl();
```
JeKachan: I think you can use testInputErrorSync helper function here.

test/testLauncher/repl.test.js
```diff
@@ -0,0 +1,139 @@
...
+		assert.ok(passed, 'Error was not thrown when it should have been.');
+		process.env.REPL_IPC_PORT = port;
+	});
+
+	it('should handle IPC startup error', async() => {
+		const replIpcErrorInChildProcess = sandbox.stub(texts, 'replIpcErrorInChildProcess');
+
+		suitestRepl.startRepl();
+
+		return new Promise(resolve => {
```
JeKachan: why this Promise needed?

lib/commands/interactive.js
```diff
@@ -0,0 +1,70 @@
...
+	const repeater = getShortFunctionBody(opts.repeater);
+
+	logger.special(replWelcomeMessage(
+		vars, cwd, watch, repeater, logger.colors.suit
+	));
+
+	isReplActive = true;
+
+	await repl.startRepl({
+		cwd,
```
JeKachan: If we run our program in node we should inherit it terminology. If `cwd` means that it will store path from directory where we run script - all ok, but if it means that cwd stores directory where script lives we should change it to `dirname`.

lib/utils/logger.js
honeydripper: I humbly disagree. 
This function does not change anything but only computes some result. In this sense I would argue that it is still pure.

Since the intent of this function is to get readable form of the currently used device, so the dependency on the environment is actually specification which is the only one that must be tested. Putting the context as default value would allow for a more ubiquitous function per se, but it would put the mentioned spec into shadow making it easy to overlook for a casual reader.

But this is all just philosophy :) @andriichumak @givehug would you also like me to change this?

lib/utils/socketChainHelper.js
honeydripper: I think these changes come from another PR and are highlighted here just because we are squashing things on merge.
I don't know what the purpose of those is anyway.

lib/utils/testHelpers/repl.js
honeydripper: Yes, it helps testing the reRequire function in repl

lib/utils/testLauncherHelper.js
honeydripper: If only you knew how many times I have changed this line during my experiments with different io methods :)
... i'll change that.

lib/utils/testLauncherHelper.js
honeydripper: This function exists for code organization only - i needed to make the calling function shorter and hide the optionality of the logStream.

But you are right it should have been marked as WriteableStream|null instead

lib/commands/interactive.js
givehug: I just realized, `this` wont work with object destructuring (`const {interactive} = sutiest;`). But importing suitest here may cause circular dependency errors. Maybe it can be safely imported into repl.js.

test/commands/interactive.test.js
honeydripper: Because I need to simulate two parallel processes calling `interactive()` in overlap, i.e. test calling `interactive()` while we are still awaiting the first `interactive()` to finish. Look closer at the location of `resolve()`.

lib/texts.js
```diff
@@ -313,6 +325,32 @@ ${leaves}`,
 	// Warnings
 	warnConfigureDeprecation: () => '.configure() command is deprecated and will be removed soon. Use .suitestrc file or cli args for configuration.',
 
+	replWarnInteractive: () => 'Ignored the "interactive" command. This command works only in the interactive mode.',
```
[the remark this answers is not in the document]
givehug: how about `.repl() , .startRepl(), .runRepl() or .debug()`?

lib/utils/logger.js
givehug: I'm ok with it, i think it does not do any argument depending computations, if it makes sense :)

test/testLauncher/repl.test.js
honeydripper: because one needs to wait until error is thrown in the uncaught exception handler, and returning promise seemed like a good idea to achieve this async test.

test/testLauncher/repl.test.js
```diff
@@ -0,0 +1,139 @@
...
+		process.env.REPL_IPC_PORT = await setupReplIpc();
+
+		suitestRepl.startRepl({
+			cwd: __dirname,
+			repeater: repeater,
+			watch: watch,
+		});
+
+		await new Promise(resolve => {
+			setInterval(() => chDir.called && resolve(), 10);
```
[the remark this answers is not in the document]
honeydripper: because there is socket communication and there is no way to know when the thing will be called.
I was actually looking for a sinon method for this. something like `stub.waitUntiCalledOrDieWithTimeout()` but could not find anything.

test/testLauncher/repl.test.js
```diff
@@ -0,0 +1,139 @@
...
+		});
+
+		await new Promise(resolve => {
+			choki.emit('change', watch);
+			setInterval(() => {
```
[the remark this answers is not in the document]
honeydripper: same as above :)

lib/commands/interactive.js
honeydripper: I think cwd in this case is accurate

lib/utils/chainPromise.js
honeydripper: @givehug could you do me a favor and check this? I don't remember much of this anymore.

lib/utils/stackTraceParser.js
```diff
@@ -83,9 +83,11 @@ function stripSuitestCodeFromStack(stack) {
 }
 
 function stripAbsolutePaths(stack) {
-	const cwdInStack = escapeRegexp('(' + process.cwd()) + '(/|\\\\)';
+	const cwdInStack = escapeRegexp(process.cwd()) + '(/|\\\\)';
```
JeKachan: I just noticed that we use process.cwd(), and I if user run suitest command somewhere outside project folder, stack trace will be generates correct?

honeydripper: in `suitest.interactive()` we change working dir to either a provided value or dirname of the file containing the call. Or do you have something else in mind?

JeKachan: I talk not only about repl mode. What if I will run full specified suitest cli command from system root (C:\ for example) for test which located somewhere in C:\User\Development\testCases\test.js, what stack trace will be generated if I will catch error in this case? Unfortunately I never check this case.

andriichumak:
@givehug @JeKachan Please someone resolve conflicts.

givehug:
fixing...

PR #137, opened 2018-10-08
SUIT-10210 Drop Symbols and Symbol mapping from JS API

AntonSBilan:
Drop Symbols and Symbol mapping from JS API


AntonSBilan:
Added ut

lib/chains/networkRequestChain.js
```diff
@@ -44,16 +44,25 @@ const toString = data => {
 		networkRequestHeaders,
 	} = texts;
 
+	const getProps = (props = []) => {
```
honeydripper: some comment would be nice

PR #151, opened 2018-11-14
Feature/suit 9655

givehug:
Add an option to explicitly set config file


andriichumak:
Would be nice to include a bit more info to PR/branch name.

PR #152, opened 2018-11-14
Feature/suit 9655 add option to set config file

givehug:
Add an option to explicitly set config file for JS API via --config cli argument.
User should be able to pass a path to any file that can be parsed with JSON.parse and override default config with it.


test/utils/timestamp.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+const cachedConfig = {...config};
+
+describe('timestamp util', () => {
+	after(() => {
+		override(cachedConfig);
+	});
+
+	it('should format date correctly', async() => {
+		override({timestamp: 'MMMM DD, YYYY'});
```
andriichumak: This sort of things make our unit tests unstable. We should rewrite API in a way that it's testable and there is no need to override global configs for this.
One option would be to rewrite "format" function to accept date format as parameter.
Other option - curry it and save in some variable.
e.g. `const format = format => date => {...}`
and then somewhere do `const formatDate = format(config.format)`. You get the idea ;)

givehug:
@andriichumak updated. I also test if helper is curried. Not sure its necessary/good practice thought.

lib/utils/timestamp.js
```diff
@@ -1,19 +1,30 @@
...
+ * Formats the date depending on timestamp in config.
+ * For logging purposes.
+ * @param {Date} date - date to format
+ * @returns {string}
+ */
+const formatDate = format(config.timestamp);
+
+module.exports = {
+	format,
+	formatDate,
```
andriichumak: So, from what I can see in this file the function that should be used in logs changed from `format` to `formatData` (as this is the one that have format pre-defined). However, I don't see corresponding change in logger... Am I missing something?

test/utils/timestamp.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+const {format} = require('../../lib/utils/timestamp');
+const timestampOff = require('../../lib/constants/timestamp').none;
+
+describe('timestamp util', () => {
+	it('should be curried', async() => {
+		const date = Date.now();
+		const fmt = 'MMMM DD, YYYY';
+
+		assert.strictEqual(format(fmt, date), format(fmt)(date), 'is curried');
```
andriichumak: I think it's irrelevant, an implementation detail that shouldn't affect behavior if changes.

lib/utils/timestamp.js
givehug: correct, thats a mistake

config/index.js
```diff
@@ -92,6 +101,19 @@ function readRcConfig() {
...
+ * Read josn config file provided by user.
+ * @param {string} configFile - path to config file
+ * @return {Object}
+ */
+function readUserConfig(configFile) {
+	// ignore config files when running unit tests
+	if (global._suitestTesting)
+		return {};
+
+	return JSON.parse(fs.readFileSync(configFile));
```
andriichumak: One last thing. Make sure that a meaningful error is shown to user e.g. if there is no such file or if JSON is not valid. Process should end with error code in such case.

givehug: Added description that it is cased by failing to process user config file. Plus original error message for more precise cause.

PR #161, opened 2018-12-19
SUIT-11598 Add appNotRunning error message

givehug:
Add error description when application is not running.


givehug:
Updated, added response.executionError handling

PR #166, opened 2019-01-09
feature/SUIT-11652 fixed error message

nivanis91:
fixed error message, when a device has been removed and test pack left with no devices

### DIFF
--- a/lib/utils/testLauncherHelper.js
+++ b/lib/utils/testLauncherHelper.js
@@ -28,6 +28,7 @@ const {mergeWith, isNil} = ramda;
 
 const {fetchLatestSuitestVersion, warnNewVersionAvailable} = require('../utils/packageMetadataHelper');
 const SuitestError = require('../utils/SuitestError');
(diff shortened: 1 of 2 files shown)


lib/utils/testLauncherHelper.js
```diff
@@ -422,8 +423,7 @@ async function prepareTestPackExecution(ownArgv) {
 	const deviceCount = devices.length;
 
 	if (!deviceCount) {
```
givehug: I wonder whether we still want to send closeSession request ?

JeKachan: It will be done in finally block in SuitestLancher

andriichumak: @nivanis91 Could you please double-check this one? If it's OK - we'll merge the PR.

nivanis91: Yep, it's done in finally block, as @JeKachan  says.

PR #169, opened 2019-01-24
SUIT-9275 Add syntax sugar for video chain subject

AntonSBilan:
Add syntax sugar for video chain subject

### DIFF
--- a/lib/chains/elementChain.js
+++ b/lib/chains/elementChain.js
@@ -278,7 +278,14 @@ const getComposers = (data) => {
 	}
 
 	if (!data.comparator && !data.isClick && !data.isMoveTo && !data.sendText && !data.isNegated) {
-		output.push(clickComposer, moveToComposer, sendTextComposer);
+		output.push(
+			clickComposer,
+			moveToComposer,
+			sendTextComposer,
+			isPlayingComposer,
+			isStoppedComposer,
+			isPausedComposer
+		);
 	}
 
 	if (data.isClick || data.sendText) {
@@ -293,10 +300,6 @@ const getComposers = (data) => {
 		if (!data.until) {
 			output.push(untilComposer);
 		}
-	}
-
-	if (!data.comparator && data.selector && data.selector.video && !data.isNegated) {
-		output.push(isPlayingComposer, isStoppedComposer, isPausedComposer);
 	}
 
 	return output;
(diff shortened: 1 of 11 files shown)


typeDefinition/VideoChain.d.ts
```diff
@@ -1,3 +1,7 @@
 import {ElementChain} from './ElementChain';
+import {VideoStateModifiers} from './modifiers';
 
-export interface VideoChain extends ElementChain {}
+export interface VideoChain extends
+	ElementChain,
+	VideoStateModifiers<VideoChain>
```
AntonSBilan: @JeKachan please review this changes

testDefinition/examplesTS/videoChain.ts
```diff
@@ -38,6 +38,10 @@ video().should.it.with.times;
 video().with.should.it.times;
 video().times.should.with.it;
 
+video().isPaused();
```
andriichumak: Add `video().isPlaying()` here as well, please

typeDefinition/VideoChain.d.ts
JeKachan: If I understnad correctly VideoStateModifiers shoud be called once, so I think that  here should be ElementChain instead of VideoChain - VideoStateModifiers<ElementChain>

AntonSBilan: You are right, thanks, but here is one more problem `video().toAssert().isPlaying()` throws error but `video().isPlaying().toAssert()` works. I could resolve it by moving  VideoStateModifiers to elementChain but in this case it will also be available for element(). Do you have any solution for this?

andriichumak: I think it's OK to allow all those modifiers on element subject as well, because you can still define video as selector and it's up to IL to figure out if it's going to work e.g. `element({css: '#video'}).isPlaying()` - should be OK.

andriichumak: At least this is how it works in UI test editor, so should be similar in JS API.

lib/chains/elementChain.js
```diff
@@ -292,6 +295,10 @@ const getComposers = (data) => {
 		}
 	}
 
+	if (!data.comparator && data.selector && data.selector.video && !data.isNegated) {
```
andriichumak: Confirmed just now that in UI Test Editor we can check video properties on any element, not only on video shortcut. So, `isPlaying`, `isPaused` and `isStopped` should be available on any element, not just the one defined as `video` selector.

typeDefinition/VideoChain.d.ts
AntonSBilan: updated

PR #170, opened 2019-01-24
Hotfix/suit 9145 - new rc/cli parameter - defaultTimeout

nivanis91:


lib/utils/testLauncherHelper.js
```diff
@@ -389,6 +389,7 @@ function getChildOptions(deviceId, deviceAccessToken, appConfigId, isDebugMode,
 		[envVars.SUITEST_APP_CONFIG_ID]: appConfigId,
 		[envVars.SUITEST_DEBUG_MODE]: isDebugMode ? 'yes' : 'no',
 		[envVars.SUITEST_CONFIG_LOG_LEVEL]: config.logLevel,
+		[envVars.SUITEST_CONFIG_DEFAULT_TIMEOUT]: config.defaultTimeout,
```
nivanis91: Having something called testLauncherHelper.js used outside test environment can cause confusion😕 I know this is a testing tool, but still...😵 maybe should be renamed to suitestLauncherHelper or something(in a new task). Just a thought.

lib/validataion/jsonSchemas.js
```diff
@@ -356,6 +356,7 @@ schemas[validationKeys.CONFIGURE] = {
 	'properties': {
 		'continueOnFatalError': {'type': 'boolean'},
 		'disallowCrashReports': {'type': 'boolean'},
+		'defaultTimeout': {}, // breaks when I put string/number, so it worked only when left empty
 		'logLevel': {'enum': ['silent', 'normal', 'verbose', 'debug']},
```
givehug: Have you tried `'type': ['number', 'string']`?

givehug: I suspect you may getting errors because same schema is used before and after this option is parsed, 'type': ['number', 'string'] might work too

config/index.js
```diff
@@ -99,9 +102,15 @@ function readRcConfig() {
  */
 function pickConfigFieldsFromEnvVars(configFields) {
 	return configFields.reduce((out, key) => {
+
 		if (ENV_VARS[key] in process.env) {
 			const val = process.env[ENV_VARS[key]];
 
+			if(isWholePositiveNumber(val)){
```
andriichumak: Are you sure about this logic? I might be paranoid, but if we really want to be precise we can't rely on format of the value to determine data type of the value. Currently it would work. What if we had other props that would fit `/\d+/`, but would still have to be a string? :/ Perhaps it's better to have a list of numeric values and do conversion only for those.

lib/validataion/jsonSchemas.js
nivanis91: Thanks, I'll give it a try

config/index.js
```diff
@@ -13,10 +13,33 @@ const {validate, validators} = require('../lib/validataion');
 const {invalidConfigObj} = require('../lib/texts');
 const {ENV_VARS} = require('../lib/mappings');
 const {pickNonNil} = require('../lib/utils/common');
+const {isWholePositiveNumber} = require('../lib/utils/stringUtils');
```
andriichumak: Is ti still used for something?

andriichumak:
Unit tests failing

config/index.js
```diff
@@ -30,6 +52,7 @@ const main = {
 	logLevel: logLevels.normal,
 	sentryDsn,
 	timestamp: timestamp.default,
+	defaultTimeout: 2000,
```
JeKachan: can be moved to the top of the file as constant

config/index.js
```diff
@@ -98,19 +123,20 @@ function readRcConfig() {
...
+
+		if (ENV_VARS[name] in process.env) {
+			const val = process.env[ENV_VARS[name]];
 
-			switch (val) {
-				case 'true':
-					out[key] = true;
+			switch (type) {
+				case 'bool':
+					out[name] = (val === 'true');
```
andriichumak: I'm curios if this wouldn't fail set-up for some people out there. E.g. from command line point of view, all of the following should yield same result:
`command --bool`
`command --bool=true`
`command --bool=1`

So, unless yargs library makes some unification there, we can't just check if it strictly equals to 'true'. Please, double-check this.

nivanis91: I double checked and yargs will parse what you provided and always return you a string as the result.
Only thing to notice is that first two scenarios will work as expected, but the third one will return you a 'false' string.
I guess that is just the way they process what we'd sent them via request.
Please let me know if we need to discuss this further.

andriichumak: In that case `command --bool=1` should be failing now as well, so it's fine. The main thing is that we are not breaking user's set-up with this change.

PR #175, opened 2019-01-29
SUIT-11909 Rename cli option config to configFile to avoid conflicts with rc module

givehug:
Rc module adds 'config' field to object it returns. This field describes path to the .rc file it founds and uses. It was conflicting with our '--config' cli option, so we decided to rename it.


andriichumak:
@givehug could you please create a ticket in Jira for this. I think we'll need to update docs as well, right?

PR #179, opened 2019-02-14
hotfix/SUIT/12032 config to configFile

nivanis91:


lib/utils/testLauncherHelper.js
```diff
@@ -395,7 +395,7 @@ function getChildOptions(deviceId, deviceAccessToken, appConfigId, isDebugMode,
 		[envVars.SUITEST_CONFIG_CONTINUE_ON_FATAL_ERROR]: config.continueOnFatalError,
 		[envVars.SUITEST_CONFIG_TIMESTAMP]: config.timestamp,
 		[envVars.SUITEST_LAUNCHER_VERSION]: version,
-		[envVars.SUITEST_USER_CONFIG]: config.config,
+		[envVars.SUITEST_USER_CONFIG]: config.configFile,
 		FORCE_COLOR: true,
```
givehug: my bad :)

AntonSBilan: maybe it worth to add some ut, which checks that correct env vars is set?

PR #184, opened 2019-02-28
Hotfix/suit 12091, 11995 refactor config sharing between processes

givehug:
SUIT-12091 fix
- Make sure yargs command handler is called only once
SUIT-11995 refactor the way config is shared between processes
- Make sure only launcher process reads config files and cli args
- Set up IPC communication between processes
- Pass config through IPC socket to children instead of env vars
- Make repl work through common IPC channel instead of its own exclusive one

IPC communication works the following way: 
- Launcher process starts ipc server, passes single env var to spawned process which includes deviceId and ipc port number
- Child process on start checks that env var, if present connects to ipc, 
- When child is connected, server sends config object to child
- Child gets config and starts session

A few commits back I had config shared as single stringified env var, so if you think this change is overkill, I can roll back to more simple solution, but that way we have data exposed in env vars.


config/index.js
```diff
@@ -6,52 +6,22 @@
...
 	'tokenKey', 'tokenPassword', 'testPackId', 'concurrency', // launcher automated
 	'username', 'password', 'orgId', 'deviceId', 'appConfigId', 'inspect', 'inspectBrk', // launcher intaractive
-	'logDir', 'timestamp', 'configFile', // launcher common
+	'logLevel', 'logDir', 'timestamp', 'configFile', 'continueOnFatalError', 'disallowCrashReports', 'defaultTimeout', // launcher common
 ];
-const allFields = [...configFields.map(({name}) => name), ...launcherFields];
 
-const main = {
+const main = Object.freeze({
 	apiUrl: 'https://the.suite.st/api/public/v2',
```
AntonSBilan: is it still v2 in docs it v3 https://suite.st/docs/suitest-network-api/api-reference/ ?

config/index.js
```diff
@@ -60,9 +30,9 @@ const main = {
...
 	defaultTimeout: DEFAULT_TIMEOUT,
 	wsUrl: 'wss://the.suite.st/api/public/v2/socket',
-};
+});
 
-const test = {
+const test = Object.freeze({
 	apiUrl: 'https://localhost',
 	continueOnFatalError: false,
 	disallowCrashReports: false,
```
AntonSBilan: should not it be true in test mode?

lib/testLauncher/ipc/server.js
```diff
@@ -0,0 +1,65 @@
...
+			);
+		}
+
+		logger.debug('Error on IPC server', err);
+	});
+
+	server.unref();
+
+	return new Promise((resolve, reject) => {
+		server.listen(0, 'localhost', err => {
```
JeKachan: server will work on localhost:0 ?

lib/utils/testLauncherHelper.js
```diff
@@ -166,15 +149,15 @@ function getDeviceLogStream(device, userArgs, startTime, logDir) {
...
+ * @returns {number|undefined}
  */
-function getDebugPort(ownArgw, runMode) {
+function getDebugPort(ownArgw, sessionType) {
 	const {inspect, inspectBrk} = ownArgw;
 	const debugPort = inspectBrk || inspect;
 
 	/* istanbul ignore if  */
-	if (debugPort && runMode === AUTOMATED) {
+	if (debugPort && sessionType === AUTOMATED) {
```
AntonSBilan: I think we need to have it well tested, Is it work with webstorm debugger? If i'm remember correctly it launches process with options.execArgv not args

lib/testLauncher/commands/interactive.js
```diff
@@ -44,29 +44,30 @@ const builder = yargs => {
 		.option('inspect', {
 			describe: 'Will launch user command with --inspect execArgv, used for debugging',
 			global: false,
+			type: 'number',
```
AntonSBilan: is it work if just pass '--inspect' without port?

AntonSBilan:
On swisstxt we have custom slack reporter and to make meaningful message there is a need of device name and manufacture. And it looks like:
`
async function getDeviceData(id) {
	const res = await fetch('https://the.suite.st/api/public/v2/devices?limit=100', {
		headers: {
			'deviceAccessToken': process.env.SUITEST_SESSION_TOKEN,
			'content-type': 'application/json'
		}
	});
	const devices = await res.json();

	return devices.values && devices.values.find(d => d.deviceId === id);
}

getDeviceData(process.env.SUITEST_DEVICE_ID).then((device) => {
	process.env.TESTING_DEVICE_NAME = `${device.manufacturer} - ${device.model}`;
	mocha.run(() => null);
});
`
Will it be possible to fetch data of device with such changes?

lib/testLauncher/ipc/server.js
AntonSBilan: I guess it will pick available one, but should we specified exactly `localhost` is it available everywhere?

config/index.js
givehug: I only know that it still works with v2, I missed it when we switched to v3, is it used in other projects already?

config/index.js
givehug: good question, I'm not sure

lib/utils/testLauncherHelper.js
givehug: well, this particular change should not break how it works now, I did not test debugging in webStorm, I thought you did recently, or @JeKachan ?

AntonSBilan: I think debugging will be working as it was but will it handle case when mode is AUTOMATED and inspect option is set not as launcher arg

lib/testLauncher/commands/interactive.js
givehug: oh, so it also has to work without it? What should happen then, webStorm will pick any available port?

AntonSBilan: it should work this way not only with webstorm but in other cases as well, it is how node works https://nodejs.org/api/debugger.html#debugger_v8_inspector_integration_for_node_js

lib/testLauncher/ipc/server.js
givehug: I think this way we actually create TCP server, not IPC :),
https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback
I'm not sure what host we should specify though.

lib/testLauncher/commands/interactive.js
givehug: ok, cool, I'll change

lib/utils/testLauncherHelper.js
givehug: How else can it be set? ownArgw in this function argument will also receive all config options passed from rc file , user config file, cli args and default config. Maybe we should rename it to be more clear.

AntonSBilan: I'm not 100% sure but webstorm and user may set it as execArgv https://nodejs.org/docs/latest/api/process.html#process_process_execargv

lib/testLauncher/commands/interactive.js
givehug: Actually good notice, in our json schema validators, type inspect and inspectBrk are set to numbers, so error is thrown if used without.

lib/utils/testLauncherHelper.js
givehug: Yeah looks like we don't handle this case

lib/testLauncher/commands/interactive.js
givehug: Another thing is that if I run script as `suitest interactive --inspect mocha ...` ownArgs.inspect is set to string 'mocha'. Maybe to consider it as true you have to set type to boolen here. I think we need to clarify this and fix as separate issue. Maybe we should force user to set some value or

andriichumak: @givehug could you please document it in a new Jira ticket, please? We'll take it into one of the following sprints.

givehug:
> Will it be possible to fetch data of device with such changes?
@AntonSBilan Actually one of the purposes of this refactoring was not to expose user data such as tokens to env vars. Another one was to make only master launcher process read all options (config files and cli args) and pass them to children. 

But session token can be accessed with `suitest.authContext.getToken()`

lib/testLauncher/ipc/ipcHelper.js
```diff
@@ -0,0 +1,68 @@
+/**
```
andriichumak: You implemented own client/server for IPC. I'm curios if it would be possible to use NodeJS built-in IPC, as described here:
https://nodejs.org/dist/latest-v11.x/docs/api/child_process.html
Is there a benefit from using raw net?

lib/testLauncher/commands/automated.js
```diff
@@ -47,20 +47,20 @@ const builder = yargs => {
...
 	await suitestLauncher.runAutomatedSession();
 };
 
+/* istanbul ignore next */
 module.exports = {
 	command,
 	describe,
 	builder,
-	handler,
+	// Make sure handler is called only once (yargs bug)
```
andriichumak: @givehug , so is it confirmed to be a bug or an assumption? If available, please put here a link to corresponding issue in their repo.

AntonSBilan:
> Device details can be accessed with `suitest.pairedDeviceContext.context`, it should already include
> manufacture and model, you don't have to make another request.
Ok I did not know this, but we started keep all device data in context not so long time ago (v1.5 October we had just deviceId) maybe we should have it in docs or better release it together with possibility provide custom reporter?

config/index.js
polomsky: Short answer: yes.
Long answer:
v2 is officially deprecated. No new features will be added to v2 and also will be removed in some time.

lib/testLauncher/commands/automated.js
givehug: https://github.com/yargs/yargs/issues/1069#issuecomment-468451450

lib/testLauncher/ipc/ipcHelper.js
givehug: Decided to implement our own, because we spawn processes.

lib/testLauncher/commands/interactive.js
givehug: https://mineus.atlassian.net/browse/SUIT-12173

PR #185, opened 2019-03-06
Hotfix/suit 11910 windows ipc fix

givehug:
Fix closing ipc connection in child process, error messages on windows.

### DIFF
--- a/lib/testLauncher/ipc/server.js
+++ b/lib/testLauncher/ipc/server.js
@@ -12,6 +12,7 @@ const {write, getState, addListener, removeListener, handleMessage, broadcast} =
 let idCount = 0;
 
 const state = {
+	server: null,
 	connections: {},
 	listeners: {},
 };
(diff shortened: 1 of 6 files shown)


lib/testLauncher/ipc/server.js
```diff
@@ -27,24 +28,27 @@ const handleMessageFromClient = handleMessage(state);
...
-		socket.on('data', handleMessageFromClient);
+		socket.on('error', err => logger.debug('IPC server socket error', err.errno));
 		socket.on('close', () => delete state.connections[socket.id]);
+		socket.on('data', handleMessageFromClient);
 		socket.unref();
 		write(socket, messageId.SETUP_SESSION, {config});
 	});
 
 	server.on('error', err => {
 		if (err.errno === 'EADDRINUSE') {
```
AntonSBilan: When we can have this message? Does it makes sense for user to know about it as there is no way user can specified port or know in advance what port will be used by launcher

givehug: I really don't know how to reproduce this error. IPC server should be started on any available port.

AntonSBilan: So maybe we can have a message that all ports are busy or node cannot find available port?

givehug: I don't know how to reproduce any of these scenarios, so I can't say which message would be better. We can try to investigate.

PR #194, opened 2019-03-19
SUIT-12239 Increase process stdout max listeners

givehug:
Explicitly increase process stdout max listeners based on number of devices (we need 2 per device) in automated test run to avoid node MaxListenersExceededWarning warning.

### DIFF
--- a/lib/testLauncher/SuitestLauncher.js
+++ b/lib/testLauncher/SuitestLauncher.js
@@ -58,8 +58,8 @@ class SuitestLauncher {
 			});
 
 			addLauncherIpcListeners();
-			// increase stdout max listeners based on number of devices to avoid node warning
-			increaseMaxListeners(process.stdout, devices.length * 2);
+			// increase stdout max listeners based on number of child processes to avoid node warning
+			increaseMaxListeners(process.stdout, devices.length, this.ownArgv.concurrency);
 
 			try {
 				await runAllDevices(
--- a/lib/testLauncher/SuitestLauncher.js
+++ b/lib/testLauncher/SuitestLauncher.js
@@ -9,6 +9,7 @@ const {
 	getDebugPort,
 	addLauncherIpcListeners,
 	throwDebugInAutomatedError,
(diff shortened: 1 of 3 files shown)


lib/testLauncher/SuitestLauncher.js
```diff
@@ -57,6 +58,8 @@ class SuitestLauncher {
 			});
 
 			addLauncherIpcListeners();
+			// increase stdout max listeners based on number of devices to avoid node warning
+			increaseMaxListeners(process.stdout, devices.length * 2);
```
AntonSBilan: we have concurrence so probably not all devices will be busier with execution at same time, and is it make sense to have the default 5 listeners as minimum value or 2 should be enough?

givehug: Default value is 10. I get current number and add 2 more per each device, because in testLauncherHelper -> process.stdout we have set 2 additional listeners to process.stdout.
It think should not matter if devices are busy or not, listeners are attached when child process is created execution.

AntonSBilan: ok, so would not it make sense to check if current limit is good before increasing it e.g. it is just 1 device so 10 is ok and we don't need 12? and I guess the point of concurrency is limit amount of cp at same time.

givehug: It would, but its just the number after which warning is displayed, you do not require extra resources with increasing number of possible listeners.

AntonSBilan: ok, thanks

andriichumak: I think it's fine, the only thing I would add is that it's not just `devices.length * 2`, but something like `(config.concurrency === 0 ? devices.length : Math.min(config.concurrency, devices.length)) * 2`

givehug: ok, i get, it doesn't make sense to add more listeners than 'concurrency' allows, I'll update

PR #195, opened 2019-03-26
SUIT-11818 Adding sendText to application command. Fixed button toString function. Fixed chainRepeat. UT improvements.

JeKachan:


test/chains/applicationChain.test.js
```diff
@@ -88,7 +141,27 @@ describe('Application chain', () => {
...
+		};
+
+		assert.strictEqual(application().toString(), 'Application has exited');
+		assert.strictEqual(sendTextApp.toString(), 'Sending text "some text" to application');
+		assert.strictEqual(sendTextApp.repeat(2).toString(), 'Sending text "some text" to application, repeat 2 times');
+		assert.strictEqual(sendTextApp.interval(2222).toString(), 'Sending text "some text" to application');
+		assert.strictEqual(
+			sendTextApp.repeat(3).interval(2222).toString(),
+			'Sending text "some text" to application, repeat 3 times every 2222 ms');
+		assert.strictEqual(sendTextApp.until(untilData).toString(), 'Sending text "some text" to application');
```
givehug: What will `sendTextApp.repeat(1).interval(1).toString()` display ?, cover those cases if there are some inconsistencies

test/chains/elementChain.test.js
```diff
@@ -197,47 +197,47 @@ describe('Element chain', () => {
 	});
 
 	it('should convert to string with meaningful message', () => {
```
givehug: I think assert.equal should be sufficient for string comparison, shouldn't it ?

JeKachan: assert.equal is deprecated - "Stability: 0 - Deprecated: Use assert.strictEqual() instead."

PR #208, opened 2019-06-24
SUIT-12786 Display console output from the device

givehug:
Display console output from the device.

### DIFF
--- a/lib/utils/logger.js
+++ b/lib/utils/logger.js
@@ -298,33 +298,40 @@ const logger = {
 				}
 			}
 		} else {
-			data.forEach(d => {
+			const args = data.map(d => {
 				if (!Array.isArray(d)) {
-					// If not an array display as it is.
-					console[method](data);
-				} else if (d[0] === 'element') {
-					// If element, diaplay as DOM el with textContent if present, do not show inner elements.
-					if (d[1].nodeType === 3) {
-						console[method](d[1].nodeValue);
-					} else {
-						const attrs = d[1].attributes
-							? Object.entries(d[1].attributes).map(([attr, val]) => ` ${attr}="${val}"`).join('')
-							: '';
-						const child = d[1].children && d[1].children[0] && d[1].children[0].nodeType === 3
-							? d[1].children[0].nodeValue
-							: '...';
-						const el = `<${d[1].nodeName}${attrs}>${child}</${d[1].nodeName}>`;
-
-						console[method](el);
-					}
-				} else if (d[0] === 'trace') {
-					// If trace, form stack frames and display each on new line.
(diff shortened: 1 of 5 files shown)


lib/utils/logger.js
```diff
@@ -246,6 +249,84 @@ const logger = {
...
+					console[method](data);
+				} else if (d[0] === 'element') {
+					// If element, diaplay as DOM el with textContent if present, do not show inner elements.
+					if (d[1].nodeType === 3) {
+						console[method](d[1].nodeValue);
+					} else {
+						const attrs = d[1].attributes
+							? Object.entries(d[1].attributes).map(([attr, val]) => ` ${attr}="${val}"`).join('')
+							: '';
+						const child = d[1].children && d[1].children[0] && d[1].children[0].nodeType === 3
```
AntonSBilan: In case of `<div>some text<span>more text</span></div>` only 'some text' is displayed and nothing shows that there is something else? and probably `d[1].children[0]` check is redundant as `children` prop presents only if there is at least one child

lib/utils/logger.js
```diff
@@ -246,6 +249,84 @@ const logger = {
...
+			}
+		} else {
+			data.forEach(d => {
+				if (!Array.isArray(d)) {
+					// If not an array display as it is.
+					console[method](data);
+				} else if (d[0] === 'element') {
+					// If element, diaplay as DOM el with textContent if present, do not show inner elements.
+					if (d[1].nodeType === 3) {
+						console[method](d[1].nodeValue);
```
AntonSBilan: maybe if it is a text node we need to wrap it into single or double quotes?

lib/utils/logger.js
givehug: We did not want to go very deep and decided to only display text if text node is there.
Open for suggestions.

AntonSBilan: Maybe we can add ... at the end if text node is to big or there is some other elements

AntonSBilan: Chrome has limit of 80 chars, if element has more text it just displays ...

lib/utils/stringUtils.js
```diff
@@ -70,11 +70,23 @@ function placeholdEmpty(s) {
 	return s === '' ? '[EMPTY]' : s;
 }
 
+/**
+ * Truncate string to 'limit' length. Append ... if over limit.
+ */
+function truncate(input, limit) {
+	if (input.length > limit) {
+		return input.substring(0, limit) + '...';
```
andriichumak: It would be cleaner to use ellipsis instead of 3 dots and substring it to `limit - 1`, so that total length including ellipsis does not exceed the limit.

There is also an edge case when we'll truncate to a space or punctuation character. Somethings like "Some text ..." or "Some text:...".

PR #210, opened 2019-07-04
SUIT-12939 Added ITEM_FOCUSED, FONT_URI to ELEMENT_PROP_TYPES. Added UT for missed types case.

JeKachan:


test/validation/elementPropTypes.test.js
```diff
@@ -0,0 +1,20 @@
+const assert = require('assert');
+
+const {ELEMENT_PROP_TYPES} = require('../../lib/validataion/elementPropTypes');
+const {ELEMENT_PROP} = require('../../lib/constants/element');
+
+describe('Testing elementPropTypes.', () => {
+	it('ELEMENT_PROP_TYPES should contain types for all ELEMENT_PROPS', () => {
+		const propsWithoutTypes = Object
+			.entries(ELEMENT_PROP)
+			.reduce((missedTypes, [key, val]) => {
```
AntonSBilan: I guess it can be 'Array.prototype.some' as if at least one does not have type test should be failed, or maybe adjust `assert.deepStrictEqual` to display what exactly is missing. But this is minor thing.

PR #213, opened 2019-07-17
SUIT-12947 Add new error code `Symbol(evaluationError)`. Display BE errorType

JeKachan:


lib/utils/socketChainHelper.js
```diff
@@ -70,7 +70,7 @@ const processServerResponse = toString => function processServerResponseHandler(
 		return false;
 	}
 	// eval fail
-	if (!isSuccess && isEval) {
+	if (!isSuccess && (isEval || (isTestLine && !isAssertionError))) {
```
givehug: Personally I think it should either be AssertionError or unknown error or something like InvalidInput error. :), but not evaluation error because its not eval chain.

andriichumak: I'm OK with invalidInput as well, as it seems to be the code we get from server. Please, discuss this with Lukas Lohnicky and then let's close the issue ;)

andriichumak:
@JeKachan please, resolve conflicts

PR #217, opened 2019-08-02
SUIT-12371 (#196)

andriichumak:
Update tvOS video states enum


andriichumak:
This is just a preparation for release, no need to review.

PR #221, opened 2019-08-07
SUIT-12786 Update device console logging

givehug:


lib/utils/logger.js
```diff
@@ -214,6 +216,47 @@ const log = method => (...messages) => {
...
+			const prefix = `${stColors.gray(timestamp.formatDate(logTime) + devicePrefix())}${stColors.bold('AL:')} `;
+
+			chunk = prefix + String(chunk).replace(/\n$/, '').replace(/\n/gm, '\n' + prefix) + '\n';
+
+			callback(null, chunk);
+		},
+	});
+
+	// Enable console colors support in node 10.16.2 and higher
+	const options = semver.gte(process.version, '10.16.2') ? [{
```
JeKachan: According to docs we can use any of these options, did not find any limitations by node versions.

givehug: We have node v8 as a requirement for js api usage in package json and for code climate we require node v9.6. Those earlier versions do not support colorMode. Code climate failed when I tried to use object as argument, thats how I noticed it.

lib/utils/logger.js
```diff
@@ -247,7 +289,8 @@ const logger = {
 	 * | ['array', any[]]
 	 * | ['element', ElementStruct]
 	 * | ['function', string]
-	 * | ['NaN']
+	 * | ['number', 'NaN' | 'Infinity' | '-Infinity']
```
AntonSBilan: So second item in array can be only  'NaN' | 'Infinity' | '-Infinity' but not a number?

PR #224, opened 2019-08-16
SUIT-13146 Added CIRCLE and CROSS to the list of buttons

andriichumak:


andriichumak:
https://mineus.atlassian.net/browse/SUIT-13146
Issue wasn't there where I was fixing it, thus no issue id in branch name

PR #227, opened 2019-08-20
SUIT-13046 - Create own videoChain implementation js-api

AntonSBilan:
separate videChain from element chain

### DIFF
--- a/lib/chains/elementChain.js
+++ b/lib/chains/elementChain.js
@@ -1,4 +1,4 @@
-const {omit, isNil, compose} = require('ramda');
+const {omit, compose} = require('ramda');
 const util = require('util');
 
 // Import utils
@@ -31,14 +31,6 @@ const {
 } = require('../composers');
 const {
 	element,
-	elementExist,
-	elementNotExist,
-	elementVisible,
-	elementMatchProps,
-	elementMatchJS,
-	elementNotMatchJS,
-	elementMatchBRS,
-	elementNotMatchBRS,
 	elementClick,
 	chainRepeat,
 	elementMoveTo,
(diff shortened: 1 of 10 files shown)


lib/chains/videoChain.js
```diff
@@ -1,8 +1,235 @@
...
+
+	return out;
+};
+const beforeSendMsg = (data) => logger.log(getOpType(data), toString(data));
+
+const toJSON = data => {
+	const type = getRequestType(data);
+	const socketMessage = {type};
+	const subject = {
+		type: 'element',
```
andriichumak: Perhaps you could also update subject definition according to https://mineus.atlassian.net/browse/SUIT-13143 - feel free to pull that ticket into sprint.
Also, should keep in mind PlayStation support release. We'll need to prepare similar changes for psVideo.

AntonSBilan: Sure, I will take this ticket

lib/chains/elementChain.js
```diff
@@ -330,6 +320,12 @@ const getComposers = (data) => {
 const elementChain = elementSelector => {
 	const selector = typeof elementSelector === 'string' ? {apiId: elementSelector} : elementSelector;
 
+	if (selector && selector.video && Object.keys(selector).length === 1) {
+		logger.warn(warnVideoAsElementDeprecation());
```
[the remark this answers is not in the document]
AntonSBilan: I'm not 100% sure we need this warning at all, as for now benefits of removing video from element probably only improving type definitions of element (remove 'isPlaying', 'isStopped', 'isPaused'), @andriichumak what do you think should it be deprecated? and if it should, what way we should use?

lib/chains/videoChain.js
```diff
@@ -1,8 +1,234 @@
...
+	// query
+	if (type === 'query') {
+		socketMessage.subject = {
+			type: 'elementProps',
+			selector: {video: true},
+		};
+	}
+
+	// video subject
+	if (data.comparator) {
```
JeKachan: this block of code (and at 69 line) are duplicates from elementChain.js. Is it possible to reuse them?

lib/chains/elementChain.js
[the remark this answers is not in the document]
andriichumak: @AntonSBilan for now that's the only benefit, but when we change selector as per SUIT-13143 - it's going to be more important.
I would go with deprecation as Vladimir described.

lib/chains/videoChain.js
andriichumak: @AntonSBilan how about moving it to some helper/decorator?

AntonSBilan: Yes, there is some duplicated lines of code, but I thought the idea of this separation is to make doing changes in element and video more clean and easier. I'm afraid that incapsulating this logic may return us back to what we had

AntonSBilan: Ok, I will update it

andriichumak: The reason we are separating this two chain subject is because they have different API. I.e. element should not have chain methods like `isPlaying`, while video should not have `setText`. So, they are different enough. However, this does not forbid us from having some code moved to helpers and reused in both cases to keep it DRY. It really depend on how much code are we talking about. If it's about 10-20 lines - probably not worth it. Otherwise I would create a helper.

PR #230, opened 2019-08-29
SUIT-13164 Do not cut stack trace in JS API when running in debug mode.

JeKachan:
Moved override config function to separate file for resolving circular dependency between config.js and SuitestError.js

### DIFF
--- a/lib/utils/stackTraceParser.js
+++ b/lib/utils/stackTraceParser.js
@@ -4,6 +4,8 @@ const escapeRegexp = require('escape-string-regexp');
 
 const isSuitestMethod = require('./isSuitestMethod');
 const {failedStackTrace} = require('../texts');
+const {config} = require('../../config');
+const logLevels = require('../constants/logLevels');
 
 /**
  * Fetch source code from stack trace frame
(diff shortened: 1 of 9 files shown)


lib/utils/stackTraceParser.js
```diff
@@ -4,6 +4,8 @@ const escapeRegexp = require('escape-string-regexp');
 
 const isSuitestMethod = require('./isSuitestMethod');
 const {failedStackTrace} = require('../texts');
+const {config} = require('../../config');
+const logLevels = require('../constants/logLevels');
```
givehug: If you decided to import config here, you can only check if logLevel is debug in stripSuitestCodeFromStack() instead of SuitestError and stackTraceWrapper

JeKachan: Yes, I thought aboit it but I decide to leave helper functions pure.

givehug: > Yes, I thought about it but I decide to leave helper functions pure.

I understand but stackTraceWrapper is also helper 8)

PR #235, opened 2019-09-06
SUIT-13296 Fix ps4 chain data json

givehug:
Fix ps4 chain data json


lib/chains/playstationVideoChain.js
```diff
@@ -73,16 +73,14 @@ const toJSON = data => {
 	const type = getRequestType(data);
 	const socketMessage = {type};
 	const subject = {
-		type: 'element',
+		type: 'psVideo',
```
JeKachan: I think you also need to fix the videoChain

PR #246, opened 2019-11-26
Feature/suit 13956 create run test chain

givehug:
Create runTest chain to be able to run test editor test by id.


lib/utils/socketChainHelper.js
```diff
@@ -114,18 +114,25 @@ const getRequestType = (data, hasQuery = true) => {
 const DEFAULT_COUNT = 1;
 const DEFAULT_DELAY = 1;
 
+const applyCount = (socketMessage, data) => {
+	socketMessage.count = 'repeat' in data ? data.repeat : DEFAULT_COUNT;
```
andriichumak: I think it would be better to have input/output immutable. But that's a minor thing, can leave it.

PR #247, opened 2019-11-27
SUIT-13423 - Config variables cannot be used and saved in non-text fi…

AntonSBilan:
allow st config vars in chains


lib/utils/socketChainHelper.js
```diff
@@ -157,8 +157,13 @@ const applyCountAndDelayAndClicks = (socketMessage, data) => {
  * @param data
  * @returns {number}
```
givehug: update this @returns {number}

lib/validation/validators.js
```diff
@@ -131,6 +131,18 @@ const validatePositiveNumber = (val, name) => {
 	return val;
 };
 
+const validateStVarOrPositiveNumber = (val, name) => {
+	if (!['number', 'string'].includes(typeof val)) {
+		throwError(name + 'should be suitest configuration variable or number');
```
givehug: missing space 'should be...'

lib/validation/validators.js
```diff
@@ -131,6 +131,18 @@ const validatePositiveNumber = (val, name) => {
...
+	} else if (typeof val === 'number' && !Number.isFinite(val) || val < 0) {
+		throwError(name + ' should be positive number');
+	} else if (typeof val === 'string' && !/(<%.+%>)/.test(val)) {
+		throwError(name + 'should be suitest configuration variable');
```
givehug: missing space 'should be...'

lib/validation/validators.js
```diff
@@ -131,6 +131,18 @@ const validatePositiveNumber = (val, name) => {
 	return val;
 };
 
+const validateStVarOrPositiveNumber = (val, name) => {
+	if (!['number', 'string'].includes(typeof val)) {
+		throwError(name + 'should be suitest configuration variable or number');
+	} else if (typeof val === 'number' && !Number.isFinite(val) || val < 0) {
+		throwError(name + ' should be positive number');
```
givehug: you can use validatePositiveNumber() here

lib/validation/validators.js
```diff
@@ -131,6 +131,18 @@ const validatePositiveNumber = (val, name) => {
 	return val;
 };
 
+const validateStVarOrPositiveNumber = (val, name) => {
```
givehug: please add UT for this function

lib/constants/validationKeys.js
```diff
@@ -3,6 +3,7 @@ const validationKeys = {
 	CONFIGURE: Symbol('configure'),
 	NON_EMPTY_STRING: Symbol('nonEmptyString'),
 	POSITIVE_NUMBER: Symbol('positiveNumber'),
+	ST_VAR_OR_POSITIVE_NUMBER: Symbol('nonStVarOrPositiveNumber'),
```
JeKachan: "non" prefix is redundant I guess?

testDefinition/examplesTS/sleepChain.ts
```diff
@@ -1,7 +1,7 @@
 import * as suitest from '../../index';
 
 async function sleepTest() {
-	const sampleSleep = suitest.sleep(10);
+	const sampleSleep = suitest.sleep(null);
```
JeKachan: sleep can accept null?

AntonSBilan: on fe it is so, one can set null by `.` to sleep line

PR #251, opened 2019-12-09
Hotfix/suit 13956 run test with line results

givehug:
Fetch test definitions in runTest chain and display output for each snippet line result. 
Requires proper line translations when ready.


andriichumak:
@JeKachan @AntonSBilan Could someone please check this branch. BE should now provide corresponding feed.

PR #260, opened 2020-01-03
SUIT-13174 Show adb response to users (#252)

andriichumak:


andriichumak:
@AntonSBilan could you please check if all good here. I created PR based on branch from Vladimir.

PR #263, opened 2020-01-10
SUIT-14177 - TypeError: appLogger[method] is not a function

AntonSBilan:


lib/utils/logger.js
```diff
@@ -251,7 +251,13 @@ function createDeviceLogger() {
 	// Return logging function
 	return (method, args, time) => {
 		logTime = time;
-		appLogger[method](...args);
+		if (appLogger[method]) {
+			appLogger[method](...args);
+		} else if (!appLogger[method] && appLogger[method.toLowerCase()]) {
+			appLogger[method.toLowerCase()](...args);
+		} else {
+			appLogger.log(...args);
```
JeKachan: what about send sentry about not supported log type from IL?
@andriichumak is it have sense?

andriichumak: sure, why not

lib/utils/logger.js
```diff
@@ -251,7 +251,13 @@ function createDeviceLogger() {
 	// Return logging function
 	return (method, args, time) => {
 		logTime = time;
-		appLogger[method](...args);
+		if (appLogger[method]) {
+			appLogger[method](...args);
+		} else if (!appLogger[method] && appLogger[method.toLowerCase()]) {
```
andriichumak: Looks weird to me, it's already an `else if`, so the `!appLogger[method]` part should be always true, as it's part of previous `if`.

Also, I think the code would be much easier to read if you do something like:
```
const logMethod = method.toLowerCase();
if (appLogger[logMethod]) {
```
No need for `else if`

AntonSBilan: there are some cammelcased methods like `timeEnd`, `countReset`,  `groupCollapsed`. The initial problem that android device can send logs method like "LOG", "ERROR"

PR #265, opened 2020-01-15
SUIT-14343 Fixed empty string processing for setText, sendText modifiers.

JeKachan:


andriichumak:
@JeKachan Please, add a unit test that covers this case for `setText` and `sendText`

PR #269, opened 2020-01-16
SUIT-14069 Point to the next version of the @suitest/translate module…

andriichumak:
…s for handling packageCorrupted startup error.


andriichumak:
@JeKachan I think the best approach would be to point to exact version, as package-lock.json would not be updated anyway. We would have to create an update in this repo when translations are updated.

PR #282, opened 2020-03-24
SUIT-14383 Add support for screenshots

JeKachan:


lib/texts.js
```diff
@@ -153,6 +153,11 @@ module.exports = {
 	location: () => 'Getting current location',
 	locationCheck: template`Checking if current location ${0} "${1}"`,
 
+	takeScreenshot: () => 'Take screenshot',
+	takeScreenshotError: () => 'Getting screenshot data failed',
```
andriichumak: Please, update text to "Failed to take screenshot"

lib/texts.js
```diff
@@ -153,6 +153,11 @@ module.exports = {
 	location: () => 'Getting current location',
 	locationCheck: template`Checking if current location ${0} "${1}"`,
 
+	takeScreenshot: () => 'Take screenshot',
```
andriichumak: Other messages here are written in present continues tense. So, should probably be "Taking screenshot"

lib/texts.js
```diff
@@ -153,6 +153,11 @@ module.exports = {
...
+	takeScreenshotError: () => 'Getting screenshot data failed',
+
+	saveScreenshot: () => 'Save screenshot',
```
andriichumak: "Saving screenshot"

package.json
```diff
@@ -71,6 +71,7 @@
   },
   "dependencies": {
     "@suitest/translate": "^2.0.5",
+    "@types/node": "^13.9.3",
```
andriichumak: Shouldn't this be in dev dependencies?

lib/utils/socketChainHelper.js
```diff
@@ -88,6 +90,27 @@ const processServerResponse = toString => function processServerResponseHandler(
...
 
+	// takeScreenshot success
+	if (isTakeScreenshot && isSuccess) {
+		if (data.dataFormat === 'raw') {
+			return res.buffer;
+		} else if (data.dataFormat === 'base64') {
+			return res.buffer.toString('base64');
+		} else if (data.fileName) {
+			try {
+				fs.writeFileSync(data.fileName, res.buffer);
```
andriichumak: Unless it's too much work, we should use async version here, so that JS thread is not blocked by filesystem operation

package.json
JeKachan: We use it for describing Buffer in .d.ts files - that why I put it to "dependecies"

lib/chains/saveScreenshotChain.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+/**
+ * @param {string} fileName
+ */
+const saveScreenshotChain = (fileName) => makeChain(getComposers, {
+	type: 'takeScreenshot',
+	fileName: validate(
+		validators.NON_EMPTY_STRING,
+		fileName,
```
andriichumak, line 46: @JeKachan we discussed file extension, I think we agreed that if file name ends with `.png` - we use it without changes. If it ends with anything else - we append `.png` to the filename.

`my-image.png` => `my-image.png`
`my-image.jpeg` => `my-image.jpeg.png`

By the way, are you sure it's always in PNG format?

andriichumak: Discussed this in Slack

PR #288, opened 2020-04-13
Feature/suit 14959 multiple sessions

AntonSBilan:
@andriichumak @JeKachan work is still in progress, created PR for progress review maybe you will have better ideas and suggestions


config/index.js
```diff
@@ -21,8 +24,8 @@ const overridableFields = [
 const main = Object.freeze({
 	apiUrl: 'https://the.suite.st/api/public/v3',
 	continueOnFatalError: false,
-	disallowCrashReports: false,
-	logLevel: logLevels.normal,
+	disallowCrashReports: true,
+	logLevel: logLevels.debug,
```
JeKachan, line 28: why log settings was changed?

AntonSBilan: dev process), will roll back with next commit, thanks

index.js
```diff
@@ -1,202 +1,15 @@
...
-			runTest: runTestAssert,
-			networkRequest: networkRequestAssert,
-			video: videoAssert,
-			psVideo: playstationVideoAssert,
-		};
-	}
-}
+const SUITEST_API = require('./suitest');
+const suitest = new SUITEST_API(
+	true, {authContext, appContext, pairedDeviceContext, testContext}
```
JeKachan, line 7: I would transform first argument to some settings object, for ability adding new options without changing constructor signature.

lib/utils/socketChainHelper.js
```diff
@@ -7,90 +7,91 @@ const {
...
-		logger.warn(infoMessage(''));
-	}
-	// fatal errors
-	if (res.result === 'fatal' && !config.continueOnFatalError) {
-		logger.error(infoMessage(''));
-		throwErr(new SuitestError(message), SuitestError.UNKNOWN_ERROR, responseForError);
-	}
-	// query
-	if (isQuery) {
+const processServerResponse = (logger, toString, continueOnFatalError) =>
```
JeKachan, line 21: cant find place where processServerResponse accepts  continueOnFatalError

lib/composers/untilComposer.js
```diff
@@ -1,20 +1,19 @@
...
-const {unregisterLeaf} = require('../utils/unusedExpressionWatchers');
 const SuitestError = require('../utils/SuitestError');
 const texts = require('../texts');
 
 /**
  * Until composer populates chain with .until() method,
  * which accepts another chain and stores it under 'until' property
  */
-const untilComposer = makeModifierComposer(composers.UNTIL, ['until'], (data, chain) => {
+const untilComposer = makeModifierComposer(composers.UNTIL, ['until'], (classInstance, data, chain) => {
```
JeKachan, line 11: classInstance only used in this place?

AntonSBilan: used mainly for expression watchers and in then chain to access websockets

lib/utils/socketChainHelper.js
AntonSBilan: cannot find it as well, @andriichumak @JeKachan is it possible that it was removed because of some BE issues?

lib/utils/SuitestError.js
```diff
@@ -34,7 +32,7 @@ class SuitestError extends Error {
 		if (response) {
 			this.response = response;
 		}
-		this.stack = logLevel === logLevels.debug ? this.stack : stripSuitestCodeFromStack(this.stack);
```
JeKachan: this logic moved to another place?

suitest.js
```diff
@@ -49,36 +49,38 @@ const {NETWORK_PROP, NETWORK_METHOD} = require('./lib/constants/networkRequest')
...
-
+const {setUpRaven} = require('./lib/utils/sentry/Raven');
 const {warnLauncherAndLibHasDiffVersions} = require('./lib/utils/packageMetadataHelper');
 const {createLogger} = require('./lib/utils/logger');
 
 // Publicly available API goes here
-class SUITEST_API {
-	constructor(exitOnError, instanceDependencies) {
-		// default instance dependencies
+class Suitest {
```
JeKachan, line 66: I think that we need use SuitestApi, or SuitestAPI beacause "suitest" library already exists and contains Suitest class

suitest.js
```diff
@@ -49,36 +49,38 @@ const {NETWORK_PROP, NETWORK_METHOD} = require('./lib/constants/networkRequest')
...
 		this.authContext = new AuthContext();
 		this.appContext = new Context();
 		this.pairedDeviceContext = new Context();
 		this.testContext = new Context();
 		this.configuration = configFactory();
 		this.config = this.configuration.config;
-		// override default by provided
-		Object.assign(this, instanceDependencies);
+		this.configuration.configurableFields.map(fieldName => {
+			this[`set${fieldName[0].toUpperCase()}${fieldName.slice(1)}`] = (val) => this.configuration.override({[fieldName]: val});
```
JeKachan, line 76: why not create single method that will accept fieldName that need to change and value? Also, need to update .d.ts file.

lib/utils/SuitestError.js
AntonSBilan: no, that's hard to do, at least I don't know the easy way, for end user it might be useless, we can import default loglevel from config file but it will be used on all instances?

suitest.js
AntonSBilan: in docs we have it like `suitest` but for me it does not meter, @andriichumak what class name should we use?

suitest.js
AntonSBilan: I added it d.ts file, Andrey sugested it, previous configure was confusing as we have other methods to set configuration, so instead of one several was introduced

lib/commands/closeSession.js
```diff
@@ -35,7 +35,8 @@ async function closeSession({webSockets, authContext, appContext, logger}) {
 			authedRequestObject,
 			() => new SuitestError(
 				authNotAllowed(closeSession.name),
-				SuitestError.AUTH_NOT_ALLOWED
+				SuitestError.AUTH_NOT_ALLOWED,
+				config.logLevel
```
JeKachan, line 39: do you know why logLevel passed here?

AntonSBilan: removed, thanks

suitest.js
JeKachan: Ok, but you defined only 4 methods for this, but configurableFileds contains 
'logLevel', 'logDir', 'timestamp', 'configFile', 'continueOnFatalError', 'disallowCrashReports', 'defaultTimeout'.
I whould define single function for all fields, or separate function for each field explicitly for reduce complexety, currently I need jump throught codebase for understanding what going on,.

AntonSBilan: Updated configurableFields, did this way to have a list of fields that user can configure during execution

suitest.js
andriichumak: I think Suitest is fine. I think @JeKachan refers to suitest test runner that some guy created long ago and already abandoned.

suitest.js
```diff
@@ -0,0 +1,209 @@
...
+		this.config = this.configuration.config;
+		this.configuration.configurableFields.map(fieldName => {
+			this[`set${fieldName[0].toUpperCase()}${fieldName.slice(1)}`] = (val) => this.configuration.override({[fieldName]: val});
+		});
+
+		// creating methods based on instance dependencies
+		this.logger = createLogger(this.config, this.pairedDeviceContext);
+		this.unusedExpressionWatchers = unusedExpressionWatchersFactory(this);
+		this.webSockets = webSocketsFactory(this);
+		setUpRaven(this.config, this.authContext);
```
andriichumak, line 83: We are setting up Raven here, before user has chance to call `setDisallowCrashReports`. Will there be an initial call to Raven right away? If I disallow crash reports, will this actually affect Raven?

suitest.js
```diff
@@ -0,0 +1,209 @@
...
+		this.IMAGE_LOAD_STATE = IMAGE_LOAD_STATE;
+		this.BORDER_STYLE = BORDER_STYLE;
+		this.VRC = VRC;
+		this.KEY = KEY;
+		this.NETWORK_PROP = NETWORK_PROP;
+		this.NETWORK_METHOD = NETWORK_METHOD;
+		this.HAD_NO_ERROR = HAD_NO_ERROR;
+
+		this.assert = {
+			application: applicationFactory(this).applicationAssert,
```
andriichumak, line 132: Not happy that factories are called twice just to get a single member from them each time. Should be called once per instance.

suitest.js
AntonSBilan: Will check if Raven does any requests on setUp, but in our set up we have 
```
		shouldSendCallback(data) {
					// send report if enabled in config and only uncaught suitest api errors
					return !config.disallowCrashReports && isSuitestUncaughtError(data.exception[0]);
				},
```
so I think if user will change disallowCrashReports latter it should prevent from sending reports

suitest.js
```diff
@@ -0,0 +1,209 @@
...
+// Publicly available API goes here
+class Suitest {
+	constructor({exitOnError}) {
+		// instance dependencies
+		this.authContext = new AuthContext();
+		this.appContext = new Context();
+		this.pairedDeviceContext = new Context();
+		this.testContext = new Context();
+		this.configuration = configFactory();
+		this.config = this.configuration.config;
```
andriichumak, line 74: Not sure if this one should be editable. Perhaps freeze it somehow.

I would make `config` a getter in this class, so that it's always returning `this.configuration.config`. Then, I would freeze `config` in config module and re-set it to a new instance each time it's overriden.

lib/commands/interactive.js
```diff
@@ -34,52 +32,55 @@ function getReplParameters(opts, suitest) {
...
-	isReplActive = false;
-	logger.info(replSessionEnded());
-}
+		isReplActive = false;
+		logger.info(replSessionEnded());
+	};
+};
 
-module.exports = interactive;
+module.exports = interactiveFactory;
```
andriichumak, line 86: I would leave this file a singleton and require it dynamically only when it's actually needed (i.e. we have interactive session started through launcher's child process).

lib/utils/socketChainHelper.js
andriichumak: Let's drop `continueOnFatalError` option

PR #292, opened 2020-04-28
SUIT-15058 - JS API incorrect .requestMatch validation

AntonSBilan:


test/composers/requestMatchesComposer.test.js
```diff
@@ -122,6 +122,24 @@ describe('Network Request Match Composer', () => {
 			},
 		});
 
+		chain.requestMatch({
+			name: NETWORK_PROP.BODY,
```
JeKachan, line 126: @AntonSBilan I think you should test METHOD here?

AntonSBilan: fixed

PR #293, opened 2020-05-04
SUIT-15079 - Remove test name argument from `startTest` command

AntonSBilan:


index.d.ts
```diff
@@ -46,7 +46,7 @@ declare namespace suitest {
 		setAppConfig(configId: string, options?: ConfigOverride): Promise<void|SuitestError>;
 		pairDevice(deviceId: string): Promise<DeviceData|SuitestError>;
 		releaseDevice(): Promise<void|SuitestError>;
-		startTest(clientTestId: string, options?: StartTestOptions): Promise<void|SuitestError>;
+		startTest(_?: string, __?: any): Promise<void|SuitestError>;
```
JeKachan, line 49: I would say to user that startTest with arguments are deprecated, for example:
`/* @deprecated use startTest instead */
startTest(clientTestId: string, options?: StartTestOptions): Promise<void|SuitestError>;
startTest(): PromiseLike<void|SuitestError>`

index.d.ts
```diff
@@ -95,7 +95,6 @@ declare namespace suitest {
 		authContext: AuthContext;
 		appContext: Context;
 		pairedDeviceContext: Context;
-		testContext: Context;
```
JeKachan, line 98: Removing `testContext` is a breaking change, we should increase suites-js-api version up to 2.0.0, or leave it as it is, but mark it as deprecated. @AntonSBilan @andriichumak

JeKachan: we dont, but user may use it, because this is part of public api.

JeKachan: I think we need just mark it as deprecated and then remove it in scope of refactoring.

AntonSBilan: I cannot find it's mentioning in docs, probably it was public just for testing

AntonSBilan: Anyway it will stop working after removing testName, so it will break the user logic if such exists

andriichumak: This was never meant for users, I think this is a mistake that we display context variables in TS definitions at all.
We can leave it here for now, but please do mark both "testContext" and `startTest` with arguments as deprecated. We will remove it when 2.0 is released.

AntonSBilan: But how should 'testContext' work, without testName there will be no data for it?

andriichumak: Keep all the logic as it is in current version of the JS API. Update only two things:
1. Add deprecation annotations to the TS and allow for not passing test name if user do not want to.
2. Make sure test name and props are not sent to server, even if user defined them and testContext is created.

lib/commands/startTest.js
```diff
@@ -29,9 +29,6 @@ async function startTest(clientTestId, {name, description} = {}) {
 		description,
 	};
 
-	// validate content json
-	validate(validators.START_TEST, content, invalidInputMessage(startTest.name, 'Start test data'));
-
 	// authorize
 	const authedContent = await authContext.authorizeWs({
 		...content,
```
JeKachan, line 34: I think you need remove `...content` for prevent sending clientTestId, name, description to server.

lib/commands/startTest.js
```diff
@@ -45,5 +42,14 @@ async function startTest(clientTestId, {name, description} = {}) {
 	testContext.setContext(content);
 	logger.info(testWasStarted(name || clientTestId));
 }
+const deprecateStartTest = util.deprecate(startTest, startTestDeprecate());
+
+function startTestWrapper(...args) {
```
JeKachan, line 47: You can check `arguments` object for displaying deprecation warning inside `startTest`

AntonSBilan: That's native way for it https://nodejs.org/dist/latest-v14.x/docs/api/util.html#util_util_deprecate_fn_msg_code and as with other places probably it should be shown just once but not every time when used

lib/commands/startTest.js
AntonSBilan: removed

PR #310, opened 2020-06-19
SUIT-14146 added translation for MoveTargetOutOfBounds error.

JeKachan:


andriichumak:
So, you decided to use custom message instead of the one in a feed?

JeKachan:
> So, you decided to use custom message instead of the one in a feed?

Yes, I decided to handle it like other selenium errors.

PR #311, opened 2020-06-24
SUIT-15336 - suitest.getPairedDevice(); command to get device info

AntonSBilan:


index.d.ts
```diff
@@ -92,6 +92,22 @@ declare namespace suitest {
 		 */
 		saveScreenshot(fileName: string): TakeScreenshotChain<void>;
 
+		getPairedDevice(): null | {
+			deviceId: string,
```
JeKachan, line 96: Did you took this information from docs?

AntonSBilan: no, from pariredDeviceContext

test/commands/getPairedDevice.test.js
```diff
@@ -0,0 +1,21 @@
+const assert = require('assert');
+const suitest = require('../../index');
+
+const obj = {};
```
JeKachan, line 4: I think that this object can be moved to test where it is used.

PR #312, opened 2020-06-25
SUIT-15442 - TypeError on displaying unused leaves warning

AntonSBilan:


test/chains/elementChain.test.js
```diff
@@ -318,6 +318,8 @@ describe('Element chain', () => {
 			element('el-api-id').setText('text string').toString(),
 			'Setting text "text string" for "el-api-id"'
 		);
+		assert.strictEqual(elementAssert({css: 'any'}).toString(), '');
```
JeKachan, line 321: I think, should be returned something similar like for not asserted command:
```javascript
suitest.element({ css: 'body' }).toString(); // Getting properties of "{"css":"body"}"
```

AntonSBilan: Did it similar to application chain (throw invalid arguments error)

lib/utils/makeChain.js
```diff
@@ -14,7 +15,11 @@ const makeChain = (getComposers, data) => {
 		return chain;
 	});
 
-	return makeNewChain(data);
+	const chain = makeNewChain(data);
+
+	registerLeaf(chain, data.stack);
```
JeKachan, line 20: If I understand correct it should cover: `suitest.element(selector)` ?

typeDefinition/modifiers.d.ts
```diff
@@ -4,6 +4,7 @@ export interface Thenable <R> {
 }
 
 export interface AbstractChain {
+	// May throw SuitestError if chain is malformed
```
JeKachan, line 7: Could you please replace it by jsdoc @throws directive?

PR #330, opened 2020-10-01
SUIT-16069 - Not display connectionNotEstablished message on command …

JeKachan:
…assertion fail.

### DIFF
--- a/lib/composers/thenComposer.js
+++ b/lib/composers/thenComposer.js
@@ -1,4 +1,5 @@
 const {makeMethodComposer} = require('../utils/makeComposer');
+const SuitestError = require('../utils/SuitestError');
 const composers = require('../constants/composer');
 const chainPromise = require('../utils/chainPromise');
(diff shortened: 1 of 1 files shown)


lib/composers/thenComposer.js
```diff
@@ -27,7 +28,9 @@ const makeThenComposer = (getSocketMessage, callback, beforeSend) => makeMethodC
 						return callback(error, data, socketMessage);
 					}))
 					.catch(err => {
```
AntonSBilan, line 30: I think it was designed to handle only error from authorizeWs that can be only AUTH_NOT_ALLOWED, did you check maybe the problem somewhere else, some Fn throws error that is not expected?

JeKachan: The point of this catch is displaying proper message if user not autorized, if some error happnes in call back function - it should be displayd properly.

AntonSBilan (approved):

PR #331, opened 2020-10-02
SUIT-13330 - Use test results translation module in JS API

AntonSBilan:


lib/utils/opType.js
```diff
@@ -29,6 +29,26 @@ module.exports = {
...
+	 * Return chain prefix A or E
+	 * @param {Object} requestObj
+	 * @returns {string}
+	 */
+	getRequestOpType(requestObj) {
+		switch (requestObj.type) {
+			case 'eval':
+				return `|${opTypes.evalLine}|`;
+			case 'query':
+				return `|${opTypes.evalLine}|`;
```
JeKachan, line 42: should be opTypes.queryLine here?

AntonSBilan: For user we have only "E" and "A" lines

andriichumak (approved):

PR #341, opened 2020-11-19
Hotfix/suit 16323 Add an option for user to get console and network logs programmatically

JeKachan:


lib/api/webSockets.js
```diff
@@ -42,6 +44,23 @@ const webSocketsFactory = ({config, logger}) => {
 
 			const content = path(['content'], obj) || {};
 
+			if (content.type === 'requestLog') {
```
andriichumak, line 47: I would also add a condition here to emit events only if there are listeners. To avoid extra processing. Some logs can be quite big.

See https://nodejs.org/api/events.html#events_emitter_listenercount_eventname

Something like:
```
if (self.listenerCount('networkLog')) { // ...
```

andriichumak (changes requested):

andriichumak (approved):

PR #342, opened 2020-11-30
Moved to NPM to unify the toolchain we are using cross repos

andriichumak:
Also bumped translation module dependency version


CONTRIBUTING.md
```diff
@@ -30,10 +30,9 @@ Check out your fork, switch to the newly created branch and start fixing the iss
 Here are the steps to get you started:
 
 * Make sure you have [NodeJS][] of v10 or higher installed.
-* Make sure you have [Yarn][] installed globally. You can use NPM instead, if you'd like.
-* Open your CLI in project's dir and run `yarn` (or `npm i`).
-* Run `yarn test-watch` (or `npm run test-watch`) to enable TDD.
-* Make sure to run `yarn lint` (or `npm run lint`) before making commit.
+* Open your CLI in project's dir and run `npm i`.
```
JeKachan, line 33: I thought that yarn supports package-lock.json. We can add explanation for yarn too

package.json
```diff
@@ -70,6 +70,7 @@
   },
   "typings": "index.d.ts",
   "devDependencies": {
+    "@types/node": "^14.0.10",
```
JeKachan, line 73: we use NodeJS namepace in declaring ISuitestBase

andriichumak: OK, I'll revert this one. But in general, it's still dev dependencies. In production JS code does not care about TypeScript interface

CONTRIBUTING.md
andriichumak: The support through import has it's down sides. I would not advertise this. People who are Yarn fans would know about this option anyway.

JeKachan (approved):

PR #347, opened 2020-12-28
SUIT-16525 - Detailed logs settings

AntonSBilan:


config/index.js
```diff
@@ -17,16 +17,24 @@ const DEFAULT_TIMEOUT = 2000;
...
 
 const main = Object.freeze({
 	apiUrl: 'https://the.suite.st/api/public/v3',
 	disallowCrashReports: false,
-	logLevel: logLevels.normal,
+	logLevelTestLines: logLevels.normal,
+	logLevelTestErrors: logLevels.normal,
+	logLevelNetworkLogs: logLevels.normal,
+	logLevelConsoleLogs: logLevels.normal,
+	logLevelTestLauncher: logLevels.normal,
```
AntonSBilan, line 36: @andriichumak that's not in ticket but I added as we have some logs that doe's not relate to any of type above and with 'logLevel': 'silent' they should not be logged

index.d.ts
```diff
@@ -64,6 +64,10 @@ declare namespace suitest {
 		setContinueOnFatalError(continueOnFatalError: ConfigureOptions['continueOnFatalError']): void;
 		setDisallowCrashReports(disallowCrashReports: ConfigureOptions['disallowCrashReports']): void;
 		setLogLevel(logLevel: ConfigureOptions['logLevel']): void;
+		setLogLevelTestLines(logLevel: ConfigureOptions['logLevel']): void;
+		setLogLevelTestErrors(logLevel: ConfigureOptions['logLevel']): void;
+		setLogLevelNetworkLogs(logLevel: ConfigureOptions['logLevel']): void;
+		setLogLevelConsoleLogs(logLevel: ConfigureOptions['logLevel']): void;
```
AntonSBilan, line 70: did not include "logLevelDebug" to be overridable

andriichumak:
@AntonSBilan , as a user I would rather define it like this in my `.suitestrc`:

```
{
  // ...
  logLevel: {
    testErrors: 'debug',
    networkLogs: 'silent',
    console.log: 'verbose',
    // ...
  }
}
```

I imagine you flatten it for CLI?

andriichumak:
Then `setLogLevel` function could also accept either a string (same level for all logs), or a hash with individual levels.

PR #350, opened 2021-02-09
Feature/SUIT-16718 not visible implementation

Noorama:


JeKachan (changes requested):
Need implement the same for video.

JeKachan (approved):

PR #353, opened 2021-02-23
SUIT-16147 Added new "screen orientations" for ios platform.

JeKachan:


index.d.ts
```diff
@@ -357,5 +357,11 @@ declare namespace suitest {
...
 	}
 
-	type ScreenOrientationValues = 'portrait' | 'portraitReversed' | 'landscape' | 'landscapeReversed';
+	type ScreenOrientationValues =
+		| 'portrait'
+		| 'portraitReversed'
+		| 'landscape'
+		| 'landscapeReversed'
+		| 'landscapeRight'
+		| 'landscapeLeft';
```
andriichumak, line 366: We should consider importing @suitest/types and exposing this type from there, but it's OK for now. Something to consider in future

andriichumak (approved):

PR #359, opened 2021-04-21
feature/SUIT-16656-close-line

Noorama:


lib/composers/closeComposer.js
```diff
@@ -0,0 +1,12 @@
+const {makeModifierComposer} = require('../utils/makeComposer');
+const composers = require('../constants/composer');
+
+/**
+ * Defined close method
+ */
+const closeComposer = makeModifierComposer(composers.CLOSE_APP, ['close'], (_, meta) => ({
+	...meta,
+	isClosed: true,
```
andriichumak, line 9: `isClosed` might be a bit confusing, without looking into context too deep I would assume that this is a check for the app being closed. I.e. like "application().hasExited()".

`isCloseApp` maybe?

lib/composers/suspendComposer.js
```diff
@@ -0,0 +1,12 @@
+const {makeModifierComposer} = require('../utils/makeComposer');
+const composers = require('../constants/composer');
+
+/**
+ * Defined suspend method
+ */
+const suspendComposer = makeModifierComposer(composers.SUSPEND_APP, ['suspend'], (_, meta) => ({
+	...meta,
+	isSuspended: true,
```
andriichumak, line 9: Same as for `isClosed`.

Maybe `isSuspendApp` or even just `suspendApp` / `closeApp` would be more clear for understanding.

andriichumak (changes requested):

andriichumak (approved):

PR #363, opened 2021-05-07
feature/SUIT-17199

Noorama:
SUIT-17199 - corresponding message was added for 403 error code. Test was added.

### DIFF
--- a/lib/texts.js
+++ b/lib/texts.js
@@ -63,7 +63,7 @@ module.exports = {
 	 * @returns {string}
 	 */
 	suitestServerError: (commandName, status, statusText) => {
-		if (status === 403 && statusText === 'Forbidden') {
+		if (status === 403) {
 			return module.exports.notWhitelistedIp();
 		}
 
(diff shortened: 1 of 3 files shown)


lib/texts.js
```diff
@@ -62,6 +63,10 @@ module.exports = {
 	 * @returns {string}
 	 */
 	suitestServerError: (commandName, status, statusText) => {
+		if (status === 403) {
+			return 'Unauthorized access - your IP address is not whitelisted. Please contact your Suitest administrator for more information.';
```
andriichumak, line 67: Do we need to duplicate the text like this, can't we cake `notWhitelistedIp` here?
Also, are you sure that every server error with 403 is about whitelisting? What about invalid token? It's also 403, right?

andriichumak (changes requested):
@Noorama please, do not duplicate text.

lib/texts.js
Noorama: Requested changes was added. Related to 403 error code: I've check for invalid token case - it is returns 401 error. But I made condition more specific - I also included statusText === 'Forbidden'.

Noorama: Seems that incorrect token can also lead to 403 - Forbidden error. In this case, what message would be better to display?

polomsky: @andriichumak @Noorama 403 is now for forbidden IP address.
In the docs it is incorrectly mentioned 403 for invalid token feed when token is invalid. So it is ok.

andriichumak (approved):

PR #368, opened 2021-05-20
Hotfix/SUIT-16525-additional-log-options

Noorama:


lib/api/webSockets.js
```diff
@@ -56,7 +56,8 @@ const webSocketsFactory = (self) => {
 				content.type === 'console' &&
 				content.data[0] !== 'SUITESTIFY: domainMapping'
 			) {
-				const {args: data, method: level} = logger.getAppOutput(content.subtype, content.data);
+				const {args: data, method: level} = logger.getAppOutput(content.subtype, content.data)
+				|| {data: undefined, level: undefined};
```
JeKachan, line 60: I think u can use here empty object `{}`

lib/api/webSockets.js
```diff
@@ -76,7 +77,7 @@ const webSocketsFactory = (self) => {
 				const requestBody = logItem.request && logItem.request.requestBody;
 				const responseBody = logItem.response && logItem.response.responseBody;
 
-				if (config.logLevel === logLevels.debug) {
+				if (config.networkLogs === logLevels.debug || config.logLevel === logLevels.debug) {
```
JeKachan, line 80: I think that specific log level should override general setting (have more weight). For example if `logLevel` defined `debug`, but `networkLogs` is not `debugg` (hidden for example) - network logs should not be displayed
@andriichumak what is preferable behavior?

andriichumak: A agree

PR #369, opened 2021-05-27
SUIT-17163 Removed redundant screen orientations

JeKachan:


andriichumak (approved):

andriichumak:
Wait a bit with merge, we would need to bump @suitest/translate as well

PR #383, opened 2021-07-08
hotfix/SUIT-17517

Noorama:
- SUIT-17517 - in case if tapType was not set, it will act as if it will have been set as single

### DIFF
--- a/lib/composers/tapComposer.js
+++ b/lib/composers/tapComposer.js
@@ -6,6 +6,7 @@ const {invalidInputMessage} = require('../texts');
 /**
  * Defines click method
  */
+const SINGLE_TAP_TYPE = 'single';
 const tapComposer = makeModifierComposer(
 	composers.TAP,
 	['tap'],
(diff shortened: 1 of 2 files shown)


lib/composers/tapComposer.js
```diff
@@ -6,14 +6,15 @@ const {invalidInputMessage} = require('../texts');
...
+const SINGLE_TAP_TYPE = 'single';
 const tapComposer = makeModifierComposer(
 	composers.TAP,
 	['tap'],
 	/**
 	 * @param {'single' | 'double' | 'long'} tapType
 	 * @param {string | number} tapDuration
 	 */
-	(_, meta, tapType, tapDuration) => {
+	(_, meta, tapType = SINGLE_TAP_TYPE, tapDuration) => {
```
JeKachan, line 17: Should be used values defined in lib/constants/tapTypes.js file

JeKachan (changes requested):

JeKachan (approved):

PR #414, opened 2021-09-21
feature/SUIT-17379

Noorama:


lib/testLauncher/composeConfig.js
```diff
@@ -45,26 +47,64 @@ const LOOKUP_DIRS = [
...
+ * @param {object} defaultConfigObject main configuration file object
+ * @param {String} extendPath extend path of other configuration file to merge with
+ * @param {String} filePath currently in read configuration file path
+ * @param {Array} foundPaths array of extends paths. Used to verify circular dependency
+ */
+function findExtendConfigs(defaultConfigObject, extendPath, filePath, foundPaths) {
+	const pathToConfig = path.join(path.dirname(filePath), extendPath);
+
+	if (foundPaths.includes(pathToConfig)) {
+		throw new SuitestError(circularDependencyError(pathToConfig), 404);
```
JeKachan, line 84: Seems that this code is rendundat here. Could you remove it pls?

test/testLauncher/composeConfig.test.js
```diff
@@ -37,5 +38,57 @@ describe('testLauncher readUserConfig', () => {
 			fs.readFileSync.restore();
 		}
 	});
+
+	it('readRcConfig should return corresponding result when argument was past', () => {
+		const CONFIG_CONTENT = '{"test": "test"}';
```
JeKachan, line 43: only module constants should be in UPPER_CASE, others should be in snakeCase

JeKachan (changes requested):

lib/testLauncher/composeConfig.js
Noorama: This code is not redundant. This part of code is responsible to check if extends have a circular recursion issue. If it is happens - it throws an error and stop further processing.

JeKachan (approved):

PR #432, opened 2021-11-01
SUIT-17639 Added suitest.relativePosition.

JeKachan:


lib/chains/relativePositionChain.js
```diff
@@ -0,0 +1,139 @@
...
+	makeToJSONComposer,
+	untilComposer,
+} = require('../composers');
+const {invalidInputMessage, relativePositionIsMalformed} = require('../texts');
+const SuitestError = require('../utils/SuitestError');
+
+/**
+ * @param {SUITEST_API} classInstance
+ */
+const relativePositionFactory = (classInstance) => {
```
polomsky, line 28: I am aware of that you are keeping existing codestyle, but I really don't know what is the purpose of it, I see no benefit:
```javascript
function relativePositionFactory(classInstance) {
const relativePositionFactory = (classInstance) => {
```

PR #434, opened 2021-11-02
SUIT-17639 Allow accepting negative numbers to relativePosition comma…

JeKachan:
…nd. Fixing typo in property on generating relativePosition chain.

### DIFF
--- a/lib/validation/validators.js
+++ b/lib/validation/validators.js
@@ -166,7 +166,7 @@ const createStVarOrNumberValidator = (onlyPositiveNumber = false) => (val, name)
 		throwError(name + ' should be suitest configuration variable or number');
 	} else if (typeof val === 'number') {
 		(onlyPositiveNumber ? validatePositiveNumber : validateNumber)(val, name);
-	} else if (typeof val === 'string' && !/(<%.+%>)/.test(val)) {
+	} else if (typeof val === 'string' && !/^<%.+%>$/.test(val)) {
 		throwError(name + ' should be suitest configuration variable');
 	}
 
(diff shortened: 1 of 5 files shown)


test/validation/validators.test.js
```diff
@@ -17,6 +19,27 @@ describe('validators', () => {
 		assert.throws(() => validatePositiveNumber(NaN), /Error/);
 	});
 
+	it('should validate number', () => {
+		assert.strictEqual(validateNumber(0), 0);
```
polomsky, line 23: negative value should be present in the test, that is the purpose of this method, add it to test

lib/validation/validators.js
```diff
@@ -153,18 +153,29 @@ const validatePositiveNumber = (val, name) => {
...
+	return val;
+};
+
+const createStVarOrNumberValidator = (onlyPositiveNumber = false) => (val, name) => {
 	if (!['number', 'string'].includes(typeof val)) {
 		throwError(name + ' should be suitest configuration variable or number');
 	} else if (typeof val === 'number') {
-		validatePositiveNumber(val, name);
+		(onlyPositiveNumber ? validatePositiveNumber : validateNumber)(val, name);
 	} else if (typeof val === 'string' && !/(<%.+%>)/.test(val)) {
```
polomsky, line 169: btw, regexp will also accept `A<%var%>B`, this is definitely wrong.
There is also `1<%var%>0` which looks very weird. Please pay attention

polomsky (changes requested):

test/validation/validators.test.js
JeKachan: added

lib/validation/validators.js
JeKachan: fixed

polomsky (approved):

PR #435, opened 2021-11-15
SUIT-18069 - getAppConfig function was added

Noorama:


lib/commands/getAppConfig.js
```diff
@@ -0,0 +1,28 @@
+const wsContentTypes = require('../api/wsContentTypes');
+const chainPromise = require('../utils/chainPromise');
+const {getAppConfigRequest, getAppConfigResponse, getAppConfigError} = require('../texts');
+
+async function getAppConfig({appContext, authContext, webSockets, logger}) {
+	logger.log(getAppConfigRequest(appContext._context.configId));
```
polomsky, line 6: is this debug code?

Noorama: No. This code is display corresponding progress message to console, while code is running.

lib/commands/getAppConfig.js
```diff
@@ -0,0 +1,28 @@
...
+		type: wsContentTypes.getAppConfig,
+		configId: appContext._context.configId,
+	}, getAppConfig.name);
+
+	const response = await webSockets.send(authedContent);
+
+	if (response.result === 'error') {
+		logger.error(getAppConfigError());
+
+		return undefined;
```
JeKachan, line 18: According to type in index.d.ts, function should no return undefined value, but shoud throw an error.

lib/commands/getAppConfig.js
JeKachan, line 6: Not sure that we should do any logging for this function - this is just utility function, that do not affect test execution.

polomsky (approved):
I think everything should be in single commit, but just saying.

JeKachan (approved):

JeKachan (approved):

PR #437, opened 2021-11-18
SUIT-18137 - false will be used in case if includeChangelist is undefined

Noorama:


polomsky:
It does not fix an issue why includeChangelist is undefined.

Noorama:
@polomsky as much as I can see, includeChangelist is something that can be defined in configs by user. It is undefined because user did not defined it or did mistake, while defining. In code side, I can completely remove ?includeChangelist={true || false} from request in case if it undefined, or assume it as false. What is preferred solution?

polomsky:
@Noorama In my opinion, the config should be normalized so that `includeChangelist` property should be defined and set to some value. Based on [docs](https://suite.st/docs/suitest-api/test-launcher/) is seems that proper value is `false` in case user will not set it up expicitly.

We will get:
- type safety for a property of config
- single point where default value can be changed for whole program

JeKachan (approved):

PR #439, opened 2021-11-26
SUIT-17689 - activeElement was added

Noorama:


suitest.js
```diff
@@ -123,6 +123,7 @@ class SUITEST_API extends EventEmitter {
 		const {saveScreenshot} = saveScreenshotFactory(this);
 		const {setScreenOrientation, setScreenOrientationAssert} = setScreenOrientationFactory(this);
 
+		this.activeElement = () => element({active: true});
```
JeKachan, line 126: Pls move this function to `elementChain.js` and add unit tests.

JeKachan (changes requested):

PR #450, opened 2022-03-11
SUIT-18459 Changed suffix for error message for fatal result.

JeKachan:


polomsky (changes requested):
This text makes no sense.

polomsky (dismissed):
I would prefer change in the single commit.

polomsky (approved):

PR #455, opened 2022-03-22
SUIT-18594-v3 Added LAUNCH_MODE to typescript definition

JeKachan:


polomsky (changes requested):
`LaunchMode` should be used in `LaunchMode` interface in modifiers.d.ts.
So probably should be also enum type.

polomsky (approved):

Issue #459, opened 2022-04-04
Type description of query results should be potentially undefined, not void


onlynone:
The type descriptions for a lot of query results are defined like:

```
type ElementQueryResult = ElementProps | void;
```

I think those should be:
```
type ElementQueryResult = ElementProps | undefined;
```

I think the type `void` is only really used for functions that don't return anything, not for an actual object.

When I use a typeguard like lodash's `isNil` on the result, it's not enough to narrow the type to just `ElementProps` since `void` isn't even `null | undefined`. A current work-around is to use `isObject`.

[The documentation](https://suite.st/docs/suitest-api-v2/chain-types/#element) even says, "If the element is not found then undefined will be returned."

I'm guessing all the instances of `someType | void` in the type definitions should be changed to something else (probably `undefined`).

polomsky:
Fixed in version 3.6.1.

PR #460, opened 2022-04-07
SUIT-18682 - Update query result types for return 'undefined' instead 'void'

JeKachan:


Noorama (approved):

typeDefinition/CookieChain.d.ts
```diff
@@ -71,5 +71,5 @@ interface CookieEvalModifiers<T> extends
 
 interface CookieAbandonedChain extends AbstractChain {}
 
-type CookieQueryResult = string | void;
+type CookieQueryResult = string | undefined;
 type CookieEvalResult = boolean | void;
```
polomsky, line 75: @JeKachan seems we have other voids

JeKachan: Making the same has no sense for boolean values. But make sense do it for keeping consistency.

polomsky (approved):

polomsky:
Fixes #459

PR #464, opened 2022-04-21
SUIT-18716-v3 Specifying "negateCondition" for "runSnippet" if "repea…

JeKachan:
…t" value greater then 1


polomsky (dismissed):
With the same note as in https://github.com/SuitestAutomation/suitest-js-api/pull/463#discussion_r855217576

polomsky (approved):

PR #467, opened 2022-06-13
SUIT-17689 - activeElement was added

Noorama:


test/chains/elementChain.test.js
```diff
@@ -906,6 +907,14 @@ describe('Element chain', () => {
 		}, 'get element css properties');
 	});
 
+	it('activeElement should work exactly as element({active: true})', () => {
+		assert.deepStrictEqual(
+			activeElement().toJSON(),
+			element({active: true}).toJSON(),
```
JeKachan, line 913: I do not see to much sense in this test - it is will be always passed even if bug will be introduced in _elementChain.js_
Pls make comparing with explicitly defined data, instead of reusing function from _elementChain.js_

JeKachan (changes requested):

JeKachan (approved):

PR #469, opened 2022-06-27
SUIT-18496 - recording was added as config option

Noorama:


JeKachan (approved):

Noorama:
Regarding ticket description changes, this PR is not actual any longer.

PR #485, opened 2022-12-15
SUIT-19461

### LINKED ISSUE #449: Using an explicit --config-file on the commandline isn't handled properly
The `composeConfig` function: https://github.com/SuitestAutomation/suitest-js-api/blob/8b09d953d3cb05c85e8aebf541b816e3b1ce8979/lib/testLauncher/composeConfig.js#L231 has some confusing behavior.

It starts off like this:
```
const composeConfig = (argv) => {
	const rcConfig = readRcConfig(argv.configFile);
	const configFilePath = argv.configFile || rcConfig.configFile;
	const userConfig = configFilePath ? readUserConfig(configFilePath) : {};

	const ownArgs = {
		...pickNonNil(suitest.configuration.overridableFields, rcConfig),
		...pickNonNil(suitest.configuration.overridableFields, userConfig),
		...pickNonNil(suitest.configuration.overridableFields, argv),
	};
```

But if you pass `--config-file` into the program, then it looks like `const rcConfig = readRcConfig(argv.configFile);` will parse that file as any of the available config file formats. And then it looks like `configFilePath` will be set to the same path, and it'll be parsed again by the line `const userConfig = configFilePath ? readUserConfig(configFilePath) : {};`. Except this time, `readUserConfig` will only try to parse the file as json:

```
function readUserConfig(path) {
	try {
		return JSON.parse(fs.readFileSync(path));
	} catch (error) {
		throw new SuitestError(invalidUserConfig(path, error.message), error.code);
	}
}
```

So you'll get an exception if your config file is in json5 or yaml or something else. Here's an example run that shows this:

```
$ ./node_modules/.bin/suitest -v
3.1.5
$ cat suitest-config.json5
{
  // This is just a test
  "tokenId": "foo",
  "tokenPassword": "bar",
}
$ ./node_modules/.bin/suitest run --config-file suitest-config.json5 mocha
SuitestError: Failed to process config file 'suitest-config.json5'.
	Unexpected token / in JSON at position 4.
	Make sure path is correct and file is in valid json format.
...
```

I think if the config file has already been parsed by `readRcConfig`, then you don't want to try parsing it again and `readUserConfig` should probably be written to accept all the different formats that `readConfigFile` accepts, no?

polomsky:
Because of `--config-file` argument produces different results than `configFile` settings, this PR deprecates both config file settings and introduces two new options:
- `--nase-config-file`
- `--override-config-file`/`overrideConfigFile`

with the clearly defined behaviour.

This PR also fixes #449.

### DIFF
--- a/lib/validation/jsonSchemas.js
+++ b/lib/validation/jsonSchemas.js
@@ -362,6 +362,7 @@ schemas[validationKeys.CONFIGURE] = {
 	'properties': {
 		'appConfigId': {'type': 'string'},
 		'concurrency': {'type': 'number'},
+		'overrideConfigFile': {'type': 'string'},
 		'configFile': {'type': 'string'},
 		'deviceId': {'type': 'string'},
 		'disallowCrashReports': {'type': 'boolean'},
--- a/lib/testLauncher/commands/run.js
+++ b/lib/testLauncher/commands/run.js
@@ -43,10 +43,21 @@ const builder = yargs => {
 			describe: 'Will launch user command with --inspect-brk execArgv, used for debugging',
 			global: false,
 		})
+		.option('override-config-file', {
+			describe: 'Specifies an additional config file which properties have precedence over properties in a standard configuration file.',
+			global: false,
+			type: 'string',
+		})
+		.option('base-config-file', {
+			describe: 'Path to a config file to override standard configuration files. Standard configuration files are ignored.',
+			global: false,
+			type: 'string',
+		})
 		.option('config-file', {
 			describe: texts.cliConfig(),
 			global: false,
 			type: 'string',
+			deprecate: 'Can have unexpected behaviour. Use --base-config-file or --override-config-file.',
 		})
 		.option('default-timeout', {
 			describe: texts.defaultTimeout(),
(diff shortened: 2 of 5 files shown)


deep9 (approved):

deep9 (approved):

deep9 (dismissed):

deep9 (dismissed):

lkrocek (approved):

lib/validation/jsonSchemas.js
```diff
@@ -362,6 +362,7 @@ schemas[validationKeys.CONFIGURE] = {
 	'properties': {
 		'appConfigId': {'type': 'string'},
 		'concurrency': {'type': 'number'},
+		'additionalConfigFile': {'type': 'string'},
```
Noorama, line 365: I think, 'singleConfigFile': {'type': 'string'} is need to be added here?

Noorama (changes requested):

lib/validation/jsonSchemas.js
polomsky: Actually no. `singleConfigFile` does not exist by purpose. In that case singleConfigFile would have same flaw as configFile.

JeKachan (approved):

PR #491, opened 2023-01-06
hotfix/SUIT-19493

Noorama:


polomsky:
@Noorama I would prefer squashing these two commits to a single one. There is not additional value in the revert commit message.

JeKachan (approved):

PR #502, opened 2023-05-04
Hotfix/SUIT-19734

polomsky:


lib/utils/fetch.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+/**
+ * Always overwrite user agent to ours
+ * @param {Object} headers
+ */
+function setUserAgent(headers) {
+	const headerName = Object.keys(headers).find(headerName => headerName.toLowerCase() === 'user-agent');
+
+	headers[headerName || 'User-Agent'] = 'Suitest JS API v' + version;
```
JeKachan, line 20: I would use common syntax for 'User-Agent': `<product>/<product-version>` where product is the same as name from package.json
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

Also I would prefer concatanate user-agent current value, instead of overriding and rename function to `addUserAgent`

JeKachan (changes requested):

lib/utils/fetch.js
polomsky: changed header generation to `suitest-js-api/{version}`

JeKachan (dismissed):

JeKachan (approved):

PR #505, opened 2023-06-07
hotfix/SUIT-19732

Noorama:


lib/api/request.js
```diff
@@ -18,19 +19,21 @@ const {captureException} = require('../utils/sentry/Raven');
...
-		if (!requestObject.headers) {
-			requestObject.headers = {};
-		}
-
 		if (!requestObject.headers['Content-Type']) {
 			requestObject.headers['Content-Type'] = 'application/json';
 		}
 	}
 
+	setUserAgent(requestObject.headers);
```
polomsky, line 35: What is the purpose of this change? `setUserAgent` is already called inside `fetch` on line [40](https://github.com/SuitestAutomation/suitest-js-api/pull/505/files#diff-7ca33c3a208a43200fedd5490d451cc48602a0282ddade8bcacedc18f5d50ab7R40)

polomsky (changes requested):

lib/api/request.js
```diff
@@ -9,6 +9,7 @@ const SuitestError = require('../utils/SuitestError');
 const makeUrlFromArray = require('../utils/makeUrlFromArray');
 const {suitestServerError} = require('../texts');
 const {captureException} = require('../utils/sentry/Raven');
+const {setUserAgent} = require('../utils/fetch');
```
polomsky, line 12: Already required on line 5. Deduplicate

polomsky (changes requested):

lib/api/request.js
Noorama: I see. So these changes are not needed at all.

PR #520, opened 2024-02-12
SUIT-9610 .acceptModal should accept empty string

andrea1gard:


lib/constants/validationKeys.js
```diff
@@ -7,7 +7,7 @@ const validationKeys = {
 	ST_VAR_OR_POSITIVE_NUMBER: Symbol('stVarOrPositiveNumberNotZero'),
 	ST_VAR_OR_NUMBER: Symbol('stVarOrNumber'),
 	NON_EMPTY_STRING_OR_UNDEFINED: Symbol('nonEmptyStringOrUndefined'),
-	NON_EMPTY_STRING_OR_NUll: Symbol('nonEmptyStringOrNull'),
+	STRING_OR_NULL: Symbol('StringOrNull'),
```
JeKachan, line 10: Symbol parameter should starts from lowecase

JeKachan (changes requested):

JeKachan (approved):

Noorama (approved):

testDefinition/examplesTS/windowChain.ts
```diff
@@ -12,6 +12,7 @@ async function windowTest() {
 	await window().goForward();
 	await window().dismissModal();
 	await window().acceptModal();
```
JeKachan, line 14: According to validation - acceptModal can't be called without arguments and may fall in runtime with validation error, please check it.

andrea1gard: sorry, just going back to this.
Do you mean in the `validateStringOrNull = (val, name)` function or somewhere else?

JeKachan: No, just need update type definition of this function

JeKachan (approved):

PR #522, opened 2024-03-07
hotfix/SUIT-20450 - compatibility of nodejs v.20 with mock-fs was partial fixed. Some part of unit-tests was skipped until nodejs mock-fs problems resolution

Noorama:


polomsky (changes requested):
I believe that skip should be conditional, for Node version 20.5 and more.

Have you tried to use a different mock system? Somebody used memfs (as shown here https://github.com/tschaub/mock-fs/issues/384)

Noorama:
> I believe that skip should be conditional, for Node version 20.5 and more.
> 
> Have you tried to use a different mock system? Somebody used memfs (as shown here [tschaub/mock-fs#384](https://github.com/tschaub/mock-fs/issues/384))

Sure, I've tried. Memfs works even worse. With mock-fs I was able to make work at least some of the tests. In addition, I don't think that these particular tests are so important to spend too much effort trying to fix them. Evgen also agrees with me.

JeKachan (approved):

PR #554, opened 2025-01-20
SUIT-21380 Added russian, ukrainian, arabic, hebrew languages

JeKachan:


Noorama (approved):

polomsky (approved):
I would prefer alphabetic order of langs

PR #558, opened 2025-02-12
hotfix/557-return-type-handle-incorrect

### LINKED ISSUE #557: Return type of handle is incorrect
`element.handle` doesn't particularly well documented (only reference I can see is https://suite.st/docs/suitest-api/assertions-test-subjects/#element so I'm not quite sure what the `multiple` flag does), but it appears the return type doesn't always match what it actually returns.

From manual testing, it seems `multiple: true` makes `handle` return an array of strings, and without it, it returns a string.

```ts
await suitest.element({active: true}).handle(); // string
await suitest.element({active: true}).handle({multiple: false}); // string
await suitest.element({active: true}).handle({multiple: true}); // string[]
```

Instead it always returns `string[]`.

aboyton:
Fixes #557

### DIFF
--- a/typeDefinition/ElementChain.d.ts
+++ b/typeDefinition/ElementChain.d.ts
@@ -125,7 +125,7 @@ interface ElementGetPropertiesChain extends
  * @description represents return value of calling "handle" function.
  */
 interface ElementHandleChain extends
-	BaseEmptyChain<ElementHandleChain, string[], ElementAbandonedChain>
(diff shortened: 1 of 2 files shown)


polomsky:
@aboyton Thank you for you PR!.

We resolved the issue as part of PR #563

polomsky:
Version 3.19.1 contains the fix.

aboyton:
Thanks for fixing this and updating the docs.

PR #577, opened 2025-09-30
 Review getPairedDevice return type

aitherios:
When registering a device it is possible to fill `Other info`, when calling `suitest.getPairedDevice` that information is supplied but it does not appear in the type.

Screenshot 2025-09-29 at 17 13 47

This patch reviews `getPairedDevice` return type by comparing the `/device/{deviceid}` endpoint documented in the [suitest api reference](https://suite.st/docs/suitest-network-api/api-reference) and filling the missing entities in the returned type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -171,13 +171,29 @@ declare namespace suitest {
 			model: string,
 			owner: string,
 			firmware: string,
+			osVersion?: string,
 			modelId: string,
 			platforms: string[],
-			customName?: string,
+			customName: string,
+			ipAddress: string,
+			controlUnitIds: string[],
 			inactivityTimeout?: number,
 			status: string,
 			displayName?: string,
 			shortDisplayName?: string,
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -178,6 +178,9 @@ declare namespace suitest {
 			status: string,
 			displayName?: string,
 			shortDisplayName?: string,
+			customUserInfo?: {
+				otherInfo: string,
```
JeKachan, line 182: Hi, thanks a lot for your contribution!

The current type definition is not entirely correct and still misses a few fields. For example, the `otherInfo` field should be defined as optional (`otherInfo?: string`) rather than required.

You can find the complete reference here:
https://suite.st/docs/suitest-network-api/api-reference/
`GET /devices/{deviceId}`

If you'd like, it would be great to add the missing fields as well.

Thanks again for your effort!

JeKachan (approved):

